PR #5, opened 2022-10-25
Build `CredentialSchema` from a JSON-schema object

lovesh:
Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/src/anonymous-credentials/schema.ts
+++ b/src/anonymous-credentials/schema.ts
@@ -3,15 +3,14 @@ import { Versioned } from './versioned';
 import { EncodeFunc, Encoder } from '../bbs-plus';
 import { isPositiveInteger } from '../util';
 import {
-  CRED_VERSION_STR,
+  CRYPTO_VERSION_STR, EMBEDDED_SCHEMA_URI_PREFIX,
   FlattenedSchema,
-  REGISTRY_ID_STR,
+  ID_STR,
   REV_CHECK_STR,
   REV_ID_STR,
-  SCHEMA_STR,
+  SCHEMA_STR, SCHEMA_TYPE_STR,
   STATUS_STR,
-  SUBJECT_STR,
-  VERSION_STR
+  SUBJECT_STR, TYPE_STR,
 } from './types-and-consts';
 import { flattenTill2ndLastKey } from './util';
 
@@ -24,7 +23,7 @@ import { flattenTill2ndLastKey } from './util';
  Some example schemas
 
  {
-  credentialVersion: {type: "string"},
+  cryptoVersion: {type: "string"},
   credentialSchema: {type: "string"},
   credentialSubject: {
     fname: {type: "string"},
@@ -44,7 +43,7 @@ import { flattenTill2ndLastKey } from './util';
  }
 
  {
-  credentialVersion: {type: "string"},
+  cryptoVersion: {type: "string"},
   credentialSchema: {type: "string"},
   credentialSubject: {
     fname: {type: "string"},
@@ -76,13 +75,13 @@ import { flattenTill2ndLastKey } from './util';
     rank: {type: "positiveInteger"}
   },
   credentialStatus: {
-    $registryId: {type: "string"},
-    $revocationCheck: {type: "string"},
-    $revocationId: {type: "string"},
+    id: {type: "string"},
+    revocationCheck: {type: "string"},
+    revocationId: {type: "string"},
   }
 
   {
-  credentialVersion: {type: "string"},
+  cryptoVersion: {type: "string"},
   credentialSchema: {type: "string"},
   credentialSubject: [
     {
@@ -119,7 +118,7 @@ import { flattenTill2ndLastKey } from './util';
  }
 
  {
-  credentialVersion: {type: "string"},
+  cryptoVersion: {type: "string"},
   credentialSchema: {type: "string"},
   credentialSubject: [
     {
@@ -166,7 +165,7 @@ import { flattenTill2ndLastKey } from './util';
 export const META_SCHEMA_STR = '$schema';
 
 export interface ISchema {
-  [CRED_VERSION_STR]: object;
+  [CRYPTO_VERSION_STR]: object;
   [SCHEMA_STR]: object;
   [SUBJECT_STR]: object | object[];
   // @ts-ignore
@@ -221,8 +220,6 @@ export type ValueTypes =
   | NumberType;
 
 export interface IJsonSchemaProperties {
-  [CRED_VERSION_STR]: object;
-  [SCHEMA_STR]: object;
   [SUBJECT_STR]: object | object[];
   // @ts-ignore
   [STATUS_STR]?: object;
@@ -262,7 +259,9 @@ export class CredentialSchema extends Versioned {
   private static readonly NUM_TYPE = 'decimalNumber';
 
   // CredentialBuilder subject/claims cannot have any of these names
-  static RESERVED_NAMES = new Set([CRED_VERSION_STR, SCHEMA_STR, SUBJECT_STR, STATUS_STR]);
+  static RESERVED_NAMES = new Set([CRYPTO_VERSION_STR, SCHEMA_STR, SUBJECT_STR, STATUS_STR]);
+
+  static IMPLICIT_FIELDS = {[CRYPTO_VERSION_STR]: {type: 'string'}, [SCHEMA_STR]: {type: 'string'}};
 
   // Custom definitions for JSON schema syntax
   static JSON_SCHEMA_CUSTOM_DEFS = {
@@ -289,9 +288,9 @@ export class CredentialSchema extends Versioned {
 
   // Keys to ignore from generic validation as they are already validated
   static IGNORE_GENERIC_VALIDATION = new Set([
-    CRED_VERSION_STR,
+    CRYPTO_VERSION_STR,
     SCHEMA_STR,
-    `${STATUS_STR}.${REGISTRY_ID_STR}`,
+    `${STATUS_STR}.${ID_STR}`,
     `${STATUS_STR}.${REV_CHECK_STR}`,
     `${STATUS_STR}.${REV_ID_STR}`
   ]);
@@ -364,6 +363,10 @@ export class CredentialSchema extends Versioned {
       encoders.set(names[i], f);
     }
 
+    // Implicitly present fields
+    encoders.set(CRYPTO_VERSION_STR, defaultEncoder);
+    encoders.set(SCHEMA_STR, defaultEncoder);
+
     // Intentionally not supplying default encoder as we already know the schema
     this.encoder = new Encoder(encoders);
   }
@@ -373,19 +376,16 @@ export class CredentialSchema extends Versioned {
    * @param schema
    */
   static validate(schema: ISchema) {
-    // Following 2 fields could have been implicit but being explicit for clarity
-    this.validateStringType(schema, CRED_VERSION_STR);
-    this.validateStringType(schema, SCHEMA_STR);
+    if (schema[SUBJECT_STR] === undefined) {
+      throw new Error(`Schema properties did not contain top level key ${SUBJECT_STR}`);
+    }
 
     const schemaStatus = schema[STATUS_STR];
     if (schemaStatus !== undefined) {
-      this.validateStringType(schemaStatus, REGISTRY_ID_STR);
+      this.validateStringType(schemaStatus, TYPE_STR);
+      this.validateStringType(schemaStatus, ID_STR);
       this.validateStringType(schemaStatus, REV_CHECK_STR);
       this.validateStringType(schemaStatus, REV_ID_STR);
-    }
-
-    if (schema[SUBJECT_STR] === undefined) {
-      throw new Error(`Schema properties did not contain top level key ${SUBJECT_STR}`);
     }
 
     this.validateGeneric(schema, CredentialSchema.IGNORE_GENERIC_VALIDATION);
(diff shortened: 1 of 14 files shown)


src/anonymous-credentials/schema.ts
```diff
@@ -587,56 +624,78 @@ export class CredentialSchema extends Versioned {
...
-        return result;
-      } else {
-        throw new Error('Schema object must have properties object');
+        case 'array':
+          return node.items.map((i) => CredentialSchema.convertToInternalSchemaObj(i, createFullName(nodeKeyName, i)));
+        default:
+          throw new Error(`Unknown type for key ${nodeKeyName} in schema: ${typ}`);
+      }
+    } else if (ref !== undefined) {
+      switch (ref) {
```
SamHellawell, line 668: This is a bad way to handle refs, refs should be defined int the schema and expanded upon parsing

tests/anonymous-credentials/schema.spec.ts
```diff
@@ -13,17 +13,118 @@ import {
...
+                  },
+                }
+              }
+            }
+          },
+        }
+      }
+    };
+    const schema1 = CredentialSchema.convertToInternalSchemaObj(jsonSchema1);
+    expect(schema1).toEqual({
```
SamHellawell, line 103: don't really see the need to convert it to another object like this if it worked by just flattening before. all this has done is remove my work and had yourself do duplicate work essentially in parsing it etc

src/anonymous-credentials/schema.ts
```diff
@@ -587,56 +624,78 @@ export class CredentialSchema extends Versioned {
...
-            }
+    // Will either have a "type" property or will be defined using "$ref"
+    const typ = node.type;
+    const ref = node.$ref;
+    if (typ !== undefined) {
+      switch (typ) {
+        case 'string':
+          return node;
+        case 'integer':
+          // If key "minimum" is not present or not an integer, validation logic in other function will throw.
```
SamHellawell, line 648: This is wrong you should add defaults incase minimum doesn't exist. I won't be forcing users to define a minimum in they certs schema creator

src/anonymous-credentials/schema.ts
```diff
@@ -587,56 +624,78 @@ export class CredentialSchema extends Versioned {
...
+          return { type: this.INT_TYPE, minimum: node.minimum };
+        case 'number':
+          return { type: this.NUM_TYPE, minimum: node.minimum, decimalPlaces: getDecimalPlaces(node.multipleOf) };
```
SamHellawell, line 651: This is wrong, what if my json schema defines number with minimum of 0? It should then internally be a positive number type

tests/anonymous-credentials/schema.spec.ts
```diff
@@ -349,7 +582,7 @@ describe('CredentialBuilder Schema', () => {
   it('subject as an array', () => {
     const schema6 = getExampleSchema(6);
     const cs6 = new CredentialSchema(schema6);
-    expect(cs6.schema[SUBJECT_STR]).toEqual(schema6[SUBJECT_STR]);
+    expect(cs6.jsonSchema.properties[SUBJECT_STR]).toEqual(schema6.properties[SUBJECT_STR]);
```
SamHellawell, line 585: this is an awful way of accessing the schema object, having var.jsonSchema.properties instead of just var.properties

src/anonymous-credentials/schema.ts
```diff
@@ -428,35 +459,41 @@ export class CredentialSchema extends Versioned {
...
-        [SCHEMA_STR]: { type: 'string' }
+        [SCHEMA_STR]: { type: 'string' },
+        [SUBJECT_STR]: {}
       }
     };
-  }
-
-  forCredential(): object {
-    return { $version: this.version, ...this.schema };
+    if (withDefinitions) {
```
SamHellawell, line 473: you shouldn't mutate the schema like this, anyway, this method will not be used by any credential creation we have in the sdk etc

src/anonymous-credentials/schema.ts
```diff
@@ -587,56 +624,78 @@ export class CredentialSchema extends Versioned {
...
+    // Will either have a "type" property or will be defined using "$ref"
+    const typ = node.type;
+    const ref = node.$ref;
+    if (typ !== undefined) {
+      switch (typ) {
+        case 'string':
+          return node;
+        case 'integer':
+          // If key "minimum" is not present or not an integer, validation logic in other function will throw.
+          return { type: this.INT_TYPE, minimum: node.minimum };
```
SamHellawell, line 649: again wrong as my internal type should be positive integer if min is greater or equal to 0

src/anonymous-credentials/schema.ts
```diff
@@ -587,56 +624,78 @@ export class CredentialSchema extends Versioned {
...
+        case '#/definitions/positiveInteger':
+          return { type: this.POSITIVE_INT_TYPE };
+        default:
+          throw new Error(`Unknown ref for key ${nodeKeyName} in schema: ${ref}`);
       }
-    } else if (node.type === 'array') {
-      return node.items.map((i) => CredentialSchema.flattenJSONSchema(i));
     } else {
-      return node;
+      if (Array.isArray(node.allOf)) {
```
SamHellawell, line 679: what is the intention here? allOf requires that a value conforms to all schemas within it - but that's not what this code and lines below are doing

src/anonymous-credentials/schema.ts
```diff
@@ -587,56 +624,78 @@ export class CredentialSchema extends Versioned {
...
-      throw new Error(`Schema node must have type field that is a string, got ${JSON.stringify(node)}`);
-    }
+  /**
+   * Convert a schema object as per JSON-schema syntax (`IJsonSchema`) to the internal representation (`ISchema`).
+   * Currently, does not check if the needed JSON-schema definitions are actually present but assumes that they will be
+   * already passed.
+   * @param node
+   * @param nodeKeyName - Name of the node, used for throwing more informative error message
+   */
+  static convertToInternalSchemaObj(node: any, nodeKeyName: string = ''): object {
```
SamHellawell, line 634: since we care about micro optimisations (assuming because you're using set over array here now)... why are we converting to an internal type then storing it? this class then has to have two member objects: internal schema and json schema, which are only set at the constructor. why? this is a waste of memory when previously we just stored one non mutated jsonschema and then parsed/flattened when required

SamHellawell (changes requested):
overall don't really understand the need for most of the changes here - and there's incorrect handling of schema types/allOf property/refs 

edit: to clarify for most of the changes, i was mainly referring to allof handling, how refs arent expanded properly and adding the jsonschema object. test and verifications are always welcome

src/anonymous-credentials/schema.ts
lovesh: > but you should not hard code them

There is no in-built type for encryptable/reversible so I don't know how else would you represent and parse it without defining a convention. Do you have an alternate in mind? 
Secondly, this was the same thing I had described when i shared the example schema in slack.

tests/anonymous-credentials/schema.spec.ts
lovesh: Benefits of converting 
1. It makes the implementation more efficient. Previously whenever schema needed to be flattened (in `flattenSchemaObj`), `flattenJSONSchema` would be called to convert the JSON-schema "style" object to another object. And schema flattening is done in both credentials and presentations. With my approach, the conversion from JSON-schema format to the one used internally is done only once while creating the `CredentialSchema` object.
2. By doing the conversion while creating `CredentialSchema` object, very little of the previous code has to change as that logic is oblivious to the fact that JSON-schema format is being used as input.

> had yourself do duplicate work essentially in parsing it etc

Much of that work is in tests which is essential. Just compare the number of checks done in your PR vs mine.

src/anonymous-credentials/schema.ts
lovesh: Before beginning this work I had mentioned in slack that error will be thrown if the expected property, `minimum` in this case, was not found. My exact quote

> The objective is for the schema creator to be very clear about what the attribute values can be. Similarly, we throw error if you did not specify multtipleOf for ...

src/anonymous-credentials/schema.ts
lovesh: Similar reasoning as above, `positiveNumber` should have been used if the caller wants to allow only positive numbers.

tests/anonymous-credentials/schema.spec.ts
lovesh: What is awful about that? `properties` no longer exists because now there are 2 schema objects and `properties` name alone can be confusing. But what is awful about that exactly?

src/anonymous-credentials/schema.ts
lovesh: Why not? I am mutating an object that i created in the same function so whats wrong with that?

src/anonymous-credentials/schema.ts
lovesh: Explained above.

src/anonymous-credentials/schema.ts
lovesh: The intention is to parse `allOf` of this form

```
"allOf": [
                { "$ref": "#/definitions/positiveNumber" },
                { "multipleOf": 0.1 }
            ]
```

If these are not found or not in the format expected, an error will be thrown while creating `CredentialSchema` obj. Any extra properties don't impact the parsing logic.

This is the exact format i had shared in slack before stating this work.

src/anonymous-credentials/schema.ts
lovesh: I explained above that the impl. of flattening is more efficient and flattening is used in both credentials and presentations. So we tradeoff memory for CPU time.

tests/anonymous-credentials/schema.spec.ts
SamHellawell: we touched on this on the call, but my initial understanding was that flattening was only called once in the constructor anyway. but i didnt look deeply into that. its not a big deal, however i disliked accessing it from outside this class as a library caller may think that mutating that value will modify the behavior/actual schema. perhaps making it private works

src/anonymous-credentials/schema.ts
SamHellawell: we discussed this on the call and can resolve it

tests/anonymous-credentials/schema.spec.ts
SamHellawell: touched on it in another comment but a library user may believe that they can modify the jsonSchema attribute to make changes, which afaik wont be the case as the constructor is parsing it into an internal representation. i think itd be better that its private and not intended for the user to access it

src/anonymous-credentials/schema.ts
SamHellawell: it feels wrong to be modifying the schema, especially for what seems an obscure method just to add definitions. i feel like it should result in the same object whether its for credential or for json encoding etc

src/anonymous-credentials/schema.ts
SamHellawell: we discussed on the call as to why its beneficial to detect this way, can resolve this thread

src/anonymous-credentials/schema.ts
SamHellawell: we discussed on the call about how allof is potentially confusing given its behaviour here is not what would be expected in jsonschema, and may not work for every possible case + since we arent relying on refs to define positive numbers (they still can if you prefer that for your schemas) as detecitng minimum value was chosen

src/anonymous-credentials/schema.ts
SamHellawell: its not worth any more changes over this as its negligable

src/anonymous-credentials/schema.ts
SamHellawell: we discussed on the call about why minimum detection is beneficial, we can still use refs by expanding them then parsing the values from that

src/anonymous-credentials/schema.ts
SamHellawell: >There is no in-built type for encryptable/reversible so I don't know how else would you represent and parse it without defining a convention. Do you have an alternate in mind?

no the definition part is fine, i have no problem with how you approached that. my concern was in ref handling (the next few lines in this method that was highlighted). please see commit below

> this was the same thing I had described when i shared the example schema in slack

yeah and to state again - no issues with that approach at all. think you misunderstood my comment. let me elaborate...

my problem related to what i mentioned with `allOf` - allowing them to use refs but only specific refs could be confusing. it makes people (potentially) believe that they can use refs in their schema - when in reality they can only use 3 hard defined refs. what i envisioned is that you would support refs using json pointers and overriding definitions (partly you did this). it wasnt too much extra effort to support refs in their entirity here.

check this commit: https://github.com/docknetwork/crypto-wasm-ts/pull/5/commits/659dd55dd74b7a18d7fcb516a706ee845eff0bec in this branch for how i envisioned it being handled. its similar to your way, but now supports expanding refs defined by the user and is a cleaner way of "hard coding" them rather than the non-modifiable switch check. now thats its a static attribute, if needed someone can modify that and also read it to understand the overriding definitions

also added test for user defined refs as they can be quite useful when defining schemas in general

tests/anonymous-credentials/schema.spec.ts
lovesh: If the only concern is unintentional changes, then it applies to a bunch of other properties across classes and we can guard that using readonly keyword.  I added that in [this commit](https://github.com/docknetwork/crypto-wasm-ts/pull/5/commits/f53bde90c48d8b2019467bf210b7270d8a3bb00a)

src/anonymous-credentials/schema.ts
lovesh: > it feels wrong to be modifying the schema, especially for what seems an obscure method just to add definitions

I am not modifying any user-given object but modifying what I created. I don't see the obscurity in creating an object based on flags passed as function arg.

> i feel like it should result in the same object whether its for credential or for json encoding etc

I agree that we can keep the custom definitions present in all cases without having any impact (except a bit of increase in   size of schema but thats not the concern) and thats the default behavior. The flag is allowing you to not have extra data in your schema when you know that you don't need to, eg. all the fields need string encoding. It's about being precise..

SamHellawell (approved):

PR #8, opened 2022-11-28
Check type before parsing

osinakayah:


lovesh (approved):

src/anonymous-credentials/credential.ts
```diff
@@ -119,7 +119,7 @@ export class Credential extends Versioned {
     });
     return new Credential(
       cryptoVersion,
-      CredentialSchema.fromJSON(JSON.parse(credentialSchema)),
+      CredentialSchema.fromJSON(typeof credentialSchema ==='string' ?  JSON.parse(credentialSchema): credentialSchema),
```
SamHellawell, line 122: @osinakayah this shouldnt be needed ideally, as the credential schema field should be a data URI json string (see examples)

PR #9, opened 2022-12-05
Support relaxed schemas

lovesh:
Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/src/anonymous-credentials/README.md
+++ b/src/anonymous-credentials/README.md
@@ -49,7 +49,6 @@ See these [tests](tests/anonymous-credentials/presentation.spec.ts) for examples
 ## TODO
 
 - Blind credential issuance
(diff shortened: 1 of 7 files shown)


SamHellawell (approved):
lgtm i think - but one thing i see that may cause issues is generating a schema based on types in the credential is that the credential issue/sign options has the schema assume "default encoding" which is strings

we either have to live with that/update SDK to generate schema on issue and realise that credentials issued w/o schema previously may be invalid if presented or assume strings/allow type detection option

i think we should be fine with just invalidating the others since its not had much use since its release

lovesh:
> realise that credentials issued w/o schema previously may be invalid

> i think we should be fine with just invalidating the others since its not had much use since its release

These will not be useful in presentations so invalidating them makes sense.

PR #10, opened 2022-12-08
Move schema generation into serializeForSigning method, add generateSchema option

SamHellawell:
Was moved here because makes more sense than mutating in sign, allows the SDK to call that method to get the same values when gathering data to sign

### DIFF
--- a/src/anonymous-credentials/credential-builder.ts
+++ b/src/anonymous-credentials/credential-builder.ts
@@ -28,7 +28,7 @@ export interface ISigningOpts {
 }
 
 export const DefaultSigningOpts: ISigningOpts = {
-  requireSameFieldsAsSchema: true
+  requireSameFieldsAsSchema: true,
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/anonymous-credentials/credential-builder.ts
```diff
@@ -171,6 +166,23 @@ export class CredentialBuilder extends Versioned {
     if (this._credStatus !== undefined) {
       s[STATUS_STR] = this._credStatus;
     }
+
+    let schema = this.schema as CredentialSchema;
```
lovesh, line 170: Why is this moved here? Previously only the sign function had to know about signing options, now serialization function also has to know about it.

lovesh: Here https://github.com/docknetwork/crypto-wasm-ts/pull/12/commits/ce3459186d444e31b694c6acf4837c7bc6196835

PR #11, opened 2022-12-10
Fix not signing proof + add proof to essential schema

SamHellawell:
The proof object was only being applied in toJSON which meant it wasnt being signed, especially noticeable with SDK credentials

### DIFF
--- a/src/anonymous-credentials/credential-builder.ts
+++ b/src/anonymous-credentials/credential-builder.ts
@@ -179,10 +179,6 @@ export class CredentialBuilder extends Versioned {
       if (signingOpts.requireSameFieldsAsSchema) {
         throw new Error('Credential does not have the fields as schema');
       } else {
-        if (schema.encoder.defaultEncoder === undefined) {
-          throw new Error('Default encoder should be defined');
-        }
-        
         // Generate new schema
         this.schema = CredentialSchema.generateAppropriateSchema(s, schema);
         schema = this.schema as CredentialSchema;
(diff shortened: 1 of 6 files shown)


src/anonymous-credentials/credential-builder.ts
```diff
@@ -167,15 +170,15 @@ export class CredentialBuilder extends Versioned {
       s[STATUS_STR] = this._credStatus;
     }
 
+    if (!s['proof']) {
```
lovesh, line 173: Proof is generated after signing so why is the "type" of proof being included in what's supposed to be signed? Or are there other details about the proof that need to be known before signing?

src/anonymous-credentials/credential.ts
```diff
@@ -12,6 +12,13 @@ import { VerifyResult } from '@docknetwork/crypto-wasm';
 import { verifyMessageObject } from '../sign-verify-js-objs';
 import b58 from 'bs58';
 
+function isEmptyObject(obj) {
```
lovesh, line 15: This is a utility function so should be moved to an appropriate file

src/anonymous-credentials/credential-builder.ts
SamHellawell: in the SDK there are other details required in proof. i would also argue that its important to sign each field in this lib's credential impl, including its type. probably makes sense to include a verification method and/or proof created dates in this impl too for signing (the SDK BBS+ credentials already have these fields and more)

lovesh: Ok.

PR #13, opened 2022-12-16
Reset presentation spec on calling finalize

SamHellawell:
Allows finalize to be called more than once, currently if it is then the credentials duplicate

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@docknetwork/crypto-wasm-ts",
-  "version": "0.31.0",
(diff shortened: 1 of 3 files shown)


lovesh:
Why aren't other fields of the builder reset when calling finalize? And why do you expect the builder to create multiple presentations from the same builder? 
Resetting some and not others is an unexpected behavior. Secondly the `finalize` method shouldn't care that it's being called again. If you need to reset the spec, then the calling code should reset whatever it needs to.

PR #15, opened 2023-03-17
Throw error when trying to encode undefined value

SamHellawell:


src/bbs-plus/encoder.ts
```diff
@@ -41,6 +41,9 @@ export class Encoder {
   encodeMessage(name: string, value: unknown, strict = false): Uint8Array {
     const encoder = this.encoders?.get(name) || this.defaultEncoder;
     if (encoder !== undefined) {
+      if (typeof value === undefined) {
```
lovesh, line 44: A check for this condition [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/util.spec.ts#L68) would be nice.

SamHellawell: added

lovesh (approved):

PR #17, opened 2023-04-24
`Coconut` integration + `BBS` integration

olegnn:


src/bbs-plus/signature.ts
```diff
@@ -167,29 +169,44 @@ export class BlindSignatureG1 extends BlindSignature {
    */
   static generate(
     commitment: Uint8Array,
-    knownMessages: Map<number, Uint8Array>,
+    revealedMessages: Map<number, Uint8Array>,
```
lovesh, line 172: Thanks for making it consistent. Please update doc

lovesh (changes requested):

src/bbs-plus/proof.ts
```diff
@@ -1,35 +1,35 @@
...
   }
 
   static initialize(
     messages: Uint8Array[],
-    signature: SignatureG1,
-    params: SignatureParamsG1,
+    signature: BBSPlusSignatureG1,
+    params: BBSPlusSignatureParamsG1,
     encodeMessages: boolean,
     blindings?: Map<number, Uint8Array>,
```
lovesh, line 30: Can you put default values for both `blindings` and `revealed` like you have in PS? Same for BBS.

src/bbs/signature.ts
```diff
@@ -0,0 +1,312 @@
...
+    return bbsVerify(messages, this.value, publicKey.value, params.value, encodeMessages);
+  }
+
+  /**
+   * Generate a request for a blind signature
+   * @param messagesToBlind - messages the requester wants to hide from the signer. The key of the map is the index of the
+   * message as per the params.
+   * @param params
+   * @param encodedMessages
+   * @param revealedMessages - Any messages that the requester wishes to inform the signer about. This is for informational
```
lovesh, line 202: Update doc

src/ps/signature.ts
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * Generate blinding for creating the commitment used in the request for blind signature
+   * @param seed - Optional seed to serve as entropy for the blinding.
+   */
+  static generateBlinding(seed?: Uint8Array): Uint8Array {
+    return generateRandomFieldElement(seed);
+  }
+
+  /**
+   * Generates a blind signature over the commitment of unknown messages and known messages
```
lovesh, line 182: Doc is incorrect.

src/ps/signature.ts
```diff
@@ -0,0 +1,366 @@
...
+   * the signers has verified the knowledge of committed messages
+   * @param knownMessages
+   * @param secretKey
+   * @param params
+   */
+  static generate(messages: Iterable<PSCommitmentOrMessage>, secretKey: PSSecretKey, h: Uint8Array): PSBlindSignature {
+    return new PSBlindSignature(psBlindSign(messages, secretKey.value, h));
+  }
+
+  /**
```
lovesh, line 193: Doc is incorrect.

src/ps/signature.ts
```diff
@@ -0,0 +1,366 @@
...
+   * Unblind the blind signature to get a regular signature that can be verified
+   * @param blinding
+   */
+  unblind(indexedBlindings: Map<number, Uint8Array>, pk: PSPublicKey): PSSignature {
```
lovesh, line 197: Doc is incorrect.

src/ps/signature.ts
```diff
@@ -0,0 +1,366 @@
...
+   * @param messagesToBlind - messages the requester wants to hide from the signer. The key of the map is the index of the
+   * message as per the params.
+   * @param blindings - If not provided, a random blindings will be generated for each message
+   * @param params
+   * @param h
+   * @param blinding
+   * @param revealedMessages - Any messages that the requester wishes to inform the signer about. This is for informational
+   * purpose only and has no cryptographic use.
+   */
+  static generateRequest(
```
lovesh, line 212: The order of args is a bit different from other 2 schemes. All blindings should follow params or come before them. Also we sometime say `revealedMessages` and sometimes `unblindedMessages`. Let's use a single term, prefer revealed as its less technical.

lovesh (changes requested):

src/ps/signature.ts
```diff
@@ -0,0 +1,366 @@
...
+    blinding: Uint8Array = generateRandomFieldElement(),
+    revealedMessages: Map<number, Uint8Array> = new Map()
+  ): [Uint8Array, PSBlindSignatureRequest] {
+    const hArr = params.getParamsForIndices([...messagesToBlind.keys()]);
+    const commitment = params.multiMessageCommitment([...messagesToBlind.values()], hArr, blinding);
+    const commitments: Map<number, Uint8Array> = new Map(
+      [...messagesToBlind.entries()]
+        .map(([idx, message]) => {
+          if (revealedMessages.has(idx)) {
+            return null as any;
```
lovesh, line 226: This should be an error as the caller is providing conflicting input.

src/bbs-plus/signature.ts
```diff
@@ -26,7 +28,7 @@ export abstract class Signature extends BytearrayWrapper {
    * @param message
    */
   static encodeMessageForSigning(message: Uint8Array): Uint8Array {
```
lovesh, line 30: Since the encoding logic is going to remain same for each sig scheme, better to move this and 3 functions below to an interface like `SigningMessageEncoder`

lovesh (changes requested):

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -764,4 +780,89 @@ export class PresentationBuilder extends Versioned {
...
+        throw new Error(`Unsupported message count: supported = ${supported}, received = ${messageCount}`);
+      }
+
+      return SetupParam.psSignaturePublicKey(pk as PSPublicKey);
+    } else {
+      throw new Error(`Public key is invalid ${pk}`);
+    }
+  }
+
+  private buildSignatureStatementFromParamsRef(
```
lovesh, line 837: This can be removed in favor of the method of the same name in `Presentation` if you pass public key instead of `signature`

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -764,4 +780,89 @@ export class PresentationBuilder extends Versioned {
...
+
+  private buildSignatureSetupParam(
+    params: SignatureParams,
+    messageCount: number
+  ): SetupParam {
+    if (params instanceof BBSSignatureParams) {
+      return SetupParam.bbsSignatureParams(params.adapt(messageCount));
+    } else if (params instanceof BBSPlusSignatureParamsG1) {
+      return SetupParam.bbsPlusSignatureParamsG1(params.adapt(messageCount));
+    } else if (params instanceof PSSignatureParams) {
```
lovesh, line 808: Redundant checks as mentioned above.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -764,4 +780,89 @@ export class PresentationBuilder extends Versioned {
...
+
+  private buildPublicKeySetupParam(
+    pk: BBSPublicKey | BBSPlusPublicKeyG2 | PSPublicKey,
+    messageCount: number
+  ): SetupParam {
+    if (pk instanceof BBSPublicKey) {
+      return SetupParam.bbsPlusSignaturePublicKeyG2(pk);
+    } else if (pk instanceof BBSPlusPublicKeyG2) {
+      return SetupParam.bbsPlusSignaturePublicKeyG2(pk);
+    } else if (pk instanceof PSPublicKey) {
```
lovesh, line 823: Redundant checks as mentioned above.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -764,4 +780,89 @@ export class PresentationBuilder extends Versioned {
...
+    signature: Signature,
+    sigParamsRef: number,
+    publicKeyRef: number,
+    revealedMessages: Map<number, Uint8Array>
+  ): Uint8Array {
+    if (signature instanceof BBSSignature) {
+      return Statement.bbsSignatureFromSetupParamRefs(sigParamsRef, publicKeyRef, revealedMessages, false);
+    } else if (signature instanceof BBSPlusSignatureG1) {
+      return Statement.bbsPlusSignatureFromSetupParamRefs(sigParamsRef, publicKeyRef, revealedMessages, false);
+    } else if (signature instanceof PSSignature) {
```
lovesh, line 847: Redundant checks as mentioned above.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -764,4 +780,89 @@ export class PresentationBuilder extends Versioned {
...
+
+  private buildWitness(
+    signature: Signature,
+    unrevealedMessages: Map<number, Uint8Array>
+  ): Uint8Array {
+    if (signature instanceof BBSSignature) {
+      return Witness.bbsSignature(signature, unrevealedMessages, false);
+    } else if (signature instanceof BBSPlusSignatureG1) {
+      return Witness.bbsPlusSignature(signature, unrevealedMessages, false);
+    } else if (signature instanceof PSSignature) {
```
lovesh, line 862: Redundant checks as mentioned above.

src/anonymous-credentials/presentation.ts
```diff
@@ -41,8 +47,12 @@ import { LegoVerifyingKey, LegoVerifyingKeyUncompressed } from '../legosnark';
 import { SaverCiphertext } from '../saver';
 import b58 from 'bs58';
 import { SetupParamsTracker } from './setup-params-tracker';
-import { PresentationBuilder } from './presentation-builder';
 import { flattenObjectToKeyValuesList } from '../util';
+import { PSPublicKey, PSSignatureParams } from '../ps';
+import { BBSPublicKey, BBSSignatureParams } from '../bbs';
+
+type SignatureParams = BBSSignatureParams | BBSPlusSignatureParamsG2 | PSSignatureParams;
```
lovesh, line 54: These union types should be moved to a place where both the builder and this class can access these

tsconfig.json
```diff
@@ -11,6 +11,7 @@
         "downlevelIteration": true,
         "baseUrl": ".",
         "esModuleInterop": true,
+        "skipLibCheck": true,
```
lovesh, line 14: What is the reason for adding this?

tests/composite-proofs/blind-signature.spec.ts
```diff
@@ -23,63 +23,70 @@ describe('Getting a blind signature, i.e. signature where signer is not aware of
...
+      [blinding, request] = BlindSignature.generateRequest(
+        blindedMessages,
+        blindings,
+        params,
+        h,
+        void 0,
+        knownMessages
+      );
+    } else if (isBBSPlus()) {
+      [blinding, request] = BlindSignature.generateRequest(blindedMessages, params, false, void 0, knownMessages);
```
lovesh, line 64: On PS doesn't support `encodeMessages` but other 2 do it should be kept true as before.

tests/composite-proofs/blind-signature.spec.ts
```diff
@@ -23,63 +23,70 @@ describe('Getting a blind signature, i.e. signature where signer is not aware of
...
+    } else {
+      request = BlindSignature.generateRequest(blindedMessages, params, false, knownMessages);
+    }
 
-    // Take parts of the sig params corresponding to the blinded messages
-    const commKey = params.getParamsForIndices(request.blindedIndices);
-    const statement1 = Statement.pedersenCommitmentG1(commKey, request.commitment);
+    if (request.blindedIndices) expect(request.blindedIndices).toEqual(blindedIndices);
```
lovesh, line 69: Better to check what the scheme is and then check that `request.blindedIndices == blindedIndices`  or keys of `request.commitments == blindedIndices`

tests/composite-proofs/blind-signature.spec.ts
```diff
@@ -90,7 +97,7 @@ describe('Getting a blind signature, i.e. signature where signer is not aware of
       messages[i] = m;
     }
 
-    const result = sig.verify(messages, pk, params, true);
+    const result = sig.verify(messages, pk, params, false);
```
lovesh, line 100: Same comment as above about `encodeMessages`

tests/composite-proofs/msg-js-obj/bbs.spec.ts
```diff
@@ -0,0 +1,589 @@
+import { initializeWasm } from '@docknetwork/crypto-wasm';
```
lovesh, line 1: This file can be removed if `scheme.ts` is used in `bbs-plus.spec.ts` in this directory.

tests/composite-proofs/bbs-signature-and-accumulator.spec.ts
```diff
@@ -1,25 +1,28 @@
...
+  KeyPair,
+  Signature,
+  SignatureParams,
+  buildStatement,
+  buildWitness,
+} from '../scheme'
 import { InMemoryState } from '../../src/accumulator/in-memory-persistence';
 
-describe('Proving knowledge of 1 BBS+ signature and a certain message in the accumulator', () => {
+describe('Proving knowledge of 1 signature and a certain message in the accumulator', () => {
```
lovesh, line 25: A general comment on this and other tests using `scheme.ts`, have the name of the sig scheme in `describe` to make it easier to detect failing tests from logs

tests/composite-proofs/msg-js-obj/ps.spec.ts
```diff
@@ -0,0 +1,589 @@
+import { initializeWasm } from '@docknetwork/crypto-wasm';
```
lovesh, line 1: Similar comment as before, this file can be removed

tests/composite-proofs/msg-js-obj/r1cs/assets-liabilities.spec.ts
```diff
@@ -316,12 +341,14 @@ describe('Proving that sum of assets is greater than sum of liabilities by 10000
 
     const witnesses = new Witnesses();
     for (let i = 0; i < numAssetCredentials; i++) {
-      witnesses.add(Witness.bbsSignature(signedAssets[i].signature, unrevealedMsgs[i], false));
+      for (const witness of [].concat(buildWitness(signedAssets[i].signature, unrevealedMsgs[i], false)))
```
lovesh, line 344: Don't need concat as `buildWitness` will only return 1 value.

tests/composite-proofs/msg-js-obj/r1cs/assets-liabilities.spec.ts
```diff
@@ -316,12 +341,14 @@ describe('Proving that sum of assets is greater than sum of liabilities by 10000
...
+        witnesses.add(witness);
     }
     for (let i = 0; i < numLiabilityCredentials; i++) {
-      witnesses.add(
-        Witness.bbsSignature(signedLiabilities[i].signature, unrevealedMsgs[numAssetCredentials + i], false)
-      );
+      for (const witness of [].concat(
```
lovesh, line 348: Same as above.

lovesh (changes requested):

src/bbs/keys.ts
```diff
@@ -0,0 +1,47 @@
+import { bbsGenerateKeyPair, bbsGeneratePublicKey } from '@docknetwork/crypto-wasm';
+import { BBSPlusPublicKeyG1, BBSPlusPublicKeyG2, BBSPlusSecretKey } from '../bbs-plus';
+import { BBSSignatureParams } from './params';
+
+/**
+ * `BBS` secret key.
+ */
+export class BBSSecretKey extends BBSPlusSecretKey {
```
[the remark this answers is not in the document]
olegnn: Makes sense because it will be possible to use it in places where  `BBSPlusSecretKey` is expected. (Same for `class BBSPublicKey extends BBSPlusPublicKeyG2`).

src/bbs/keys.ts
```diff
@@ -0,0 +1,47 @@
...
+  generatePublicKeyG1(_): BBSPlusPublicKeyG1 {
```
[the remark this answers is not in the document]
olegnn: This is inherited from `BBSPlusSecretKey`

src/bbs/keys.ts
```diff
@@ -0,0 +1,47 @@
...
+    throw new Error('Not supported')
+  }
+
+  generatePublicKey(params: BBSSignatureParams): BBSPublicKey {
```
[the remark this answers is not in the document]
olegnn: We don't have a key for `BBS` in `G1` either. Currently, `generatePublicKeyG2` allows generating a `PublicKey` using `BBSPlus` params

src/bbs-plus/params.ts
```diff
@@ -133,42 +143,127 @@ export class SignatureParamsG1 extends SignatureParams {
...
+    const sigParams = this.getSigParamsOfRequiredSize(msgCount, labelOrParams);
+    return signature.verify(encodedValues, publicKey, sigParams, false);
+  }
+
+  /**
+  * Gives `SignatureParamsG1` that can sign `msgCount` number of messages.
+  * @param msgCount
+  * @param labelOrParams
+  */
+  static getSigParamsOfRequiredSize(
```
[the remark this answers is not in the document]
olegnn: The problem is with `labelOrParams: Uint8Array | BBSSignatureParams`, so we should know what type must be constructed in case a label is provided.
But I think about recreating `ISignatureParams` as an abstract class and moving there all repeated methods implementation.

tsconfig.json
olegnn:
```
$ tsc
node_modules/@types/node/globals.d.ts:72:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; timeout(milliseconds: number): AbortSignal; }'.

72 declare var AbortSignal: {
               ~~~~~~~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:2090:13
    2090 declare var AbortSignal: {
                     ~~~~~~~~~~~
    'AbortSignal' was also declared here.

../crypto-wasm/lib/type_declarations/index.d.ts:3:1 - error TS2308: Module "./bbs_plus" has already exported a member named 'DEFAULT_BLS12381_G1_PUBLIC_KEY_LENGTH'. Consider explicitly re-exporting to resolve the ambiguity.

3 export * from "./bbs";
  ~~~~~~~~~~~~~~~~~~~~~~

../crypto-wasm/lib/type_declarations/index.d.ts:3:1 - error TS2308: Module "./bbs_plus" has already exported a member named 'DEFAULT_BLS12381_G2_PUBLIC_KEY_LENGTH'. Consider explicitly re-exporting to resolve the ambiguity.

3 export * from "./bbs";
  ~~~~~~~~~~~~~~~~~~~~~~

../crypto-wasm/lib/type_declarations/index.d.ts:3:1 - error TS2308: Module "./bbs_plus" has already exported a member named 'DEFAULT_BLS12381_PRIVATE_KEY_LENGTH'. Consider explicitly re-exporting to resolve the ambiguity.

3 export * from "./bbs";
  ~~~~~~~~~~~~~~~~~~~~~~


Found 4 errors in 2 files.

Errors  Files
     1  node_modules/@types/node/globals.d.ts:72
     3  ../crypto-wasm/lib/type_declarations/index.d.ts:3
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -308,22 +319,19 @@ export class PresentationBuilder extends Versioned {
...
-    // Create statements and witnesses for proving possession of each credential, i.e. proof of knowledge of BBS+ sigs.
+    // Create statements and witnesses for proving possession of each credential, i.e. proof of knowledge of the sigs.
     // Also collect encoded attributes used in any predicate
     for (let i = 0; i < numCreds; i++) {
       const cred = this.credentials[i][0];
       const schema = cred.schema as CredentialSchema;
       const flattenedSchema = schema.flatten();
 
       const numAttribs = flattenedSchema[0].length;
-      if (maxAttribs < numAttribs) {
```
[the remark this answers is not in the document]
olegnn: As a performance improvement, we can adapt the last used params if they have the same type as current

tsconfig.json
olegnn: I'll update exported constants in `crypto-wasm-ts` soon

olegnn: https://github.com/docknetwork/crypto-wasm/pull/8

src/bbs-plus/signature.ts
```diff
@@ -155,41 +73,56 @@ export abstract class BlindSignature extends BytearrayWrapper {
...
         } must be less than ${params.supportedMessageCount()} supported by the signature params`
       );
     }
-    const sig = bbsBlindSignG1(commitment, knownMessages, secretKey.value, params.value, encodeMessages);
-    return new BlindSignatureG1(sig);
+    const sig = bbsPlusBlindSignG1(commitment, revealedMessages, secretKey.value, params.value, encodeMessages);
+    return new BBSPlusBlindSignatureG1(sig);
+  }
+
+  /**
```
lovesh, line 104: Update doc.

src/bbs-plus/signature.ts
```diff
@@ -204,26 +137,65 @@ export class BlindSignatureG1 extends BlindSignature {
    */
   static generateRequest(
     messagesToBlind: Map<number, Uint8Array>,
-    params: SignatureParamsG1,
+    params: BBSPlusSignatureParamsG1,
     encodeMessages: boolean,
     blinding?: Uint8Array,
     unblindedMessages?: Map<number, Uint8Array>
```
lovesh, line 143: `unblindedMessages` -> `revealedMessages`

src/bbs/signature.ts
```diff
@@ -0,0 +1,212 @@
...
+      encodedMessages,
+      signature: blindSig
+    };
+  }
+
+  /**
+   * Generate a blind signature from request
+   * @param request
+   * @param secretKey
+   * @param h
```
lovesh, line 181: Update doc

lovesh (changes requested):

src/anonymous-credentials/types-and-consts.ts
```diff
@@ -57,3 +83,40 @@ export const ACCUMULATOR_PROVING_KEY_LABEL_BYTES = te.encode(ACCUMULATOR_PROVING
...
+  return function () {
+    if (calculated === NO_VALUE) {
+      return (calculated = f());
+    } else {
+      return calculated;
+    }
+  };
+}
+
+export const DEFAULT_SIGNATURE_LABEL_BYTES = Object.setPrototypeOf(
```
lovesh, line 100: Why not just 
```
DEFAULT_SIGNATURE_LABEL_BYTES = { 
   [BBSSignatureParams.name]: BBS_SIGNATURE_PARAMS_LABEL_BYTES , 
  ..
}
```

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -304,26 +314,44 @@ export class PresentationBuilder extends Versioned {
...
       let revealedNames = this.revealedAttributes.get(i);
       if (revealedNames === undefined) {
         revealedNames = new Set();
       }
+      const paramsClass = paramsClassBySignature(cred.signature);
+      if (paramsClass == null) {
+        throw new Error(`Invalid signature: ${cred.signature}`);
+      }
+
+      sigParamsByScheme[paramsClass.name] ??= {
```
lovesh, line 339: Why not pass `num` as `numAttribs` at this point?

src/anonymous-credentials/types-and-consts.ts
```diff
@@ -57,3 +83,40 @@ export const ACCUMULATOR_PROVING_KEY_LABEL_BYTES = te.encode(ACCUMULATOR_PROVING
...
+    [BBSPlusSignatureParamsG1.name]: BBS_PLUS_SIGNATURE_PARAMS_LABEL_BYTES,
+    [PSSignatureParams.name]: PS_SIGNATURE_PARAMS_LABEL_BYTES
+  },
+  null
+);
+
+export const DEFAULT_SIGNATURE_PARAMS = Object.setPrototypeOf(
+  {
+    [BBSSignatureParams.name]: once(() =>
+      BBSSignatureParams.generate(2, DEFAULT_SIGNATURE_LABEL_BYTES[BBSSignatureParams.name])
```
lovesh, line 112: This is always going to be wasteful as all creds have more than 2 attributes

src/anonymous-credentials/presentation.ts
```diff
@@ -113,15 +113,34 @@ export class Presentation extends Versioned {
 
       const revealedEncoded = Presentation.encodeRevealed(i, presentedCred, presentedCredSchema, flattenedSchema[0]);
 
-      if (maxAttribs < numAttribs) {
-        sigParams = sigParams.adapt(numAttribs);
-        maxAttribs = numAttribs;
+      const paramsClass = paramsClassByPublicKey(publicKeys[i]);
```
lovesh, line 116: This logic is duplicated in` presentation-builder.ts` as well.

lovesh (changes requested):

src/anonymous-credentials/types-and-consts.ts
olegnn: I don't understand your point because the same thing happens at the current code https://github.com/docknetwork/crypto-wasm-ts/blob/master/src/anonymous-credentials/presentation-builder.ts#L291.
The current code is better because these parameters will be initialized once while the previous code recreated them every time.
Although we may define another amount of parameters.

src/anonymous-credentials/types-and-consts.ts
olegnn: Because of the object prototype properties.

src/anonymous-credentials/presentation-builder.ts
olegnn: Not sure what you mean. This is done this way to avoid regenerating the same default parameters for each schema on every call to this function.

src/anonymous-credentials/presentation.ts
olegnn: Well, this 224-line `verify` function has a lot of common code with the 437-line function `finalize` from the `presentation-builder`. I guess it might be refactored in a separate task as this doesn't affect anything.

src/anonymous-credentials/types-and-consts.ts
lovesh: Can you please elaborate?

src/anonymous-credentials/presentation-builder.ts
lovesh: I understand but you can generate the params for `numAttribs` and not `2`

src/anonymous-credentials/types-and-consts.ts
lovesh: The existing code that you referenced is wastefull as well. `sigParams` (of corresponding scheme) should be unset at the beginning and should be set the first time it needs to be used and should be created with `numAttribs` (in that context)

src/anonymous-credentials/presentation.ts
lovesh: I mentioned this because it looked simple to abstract in a function that takes `sigParamsByScheme` and `paramsClass`

src/anonymous-credentials/presentation-builder.ts
olegnn: Let's use your approach then.
When I wrote this, I guessed that it would be less wasteful to extend existing `n`-message signature params by `m` more messages than generating new params for `n + m` messages.

src/anonymous-credentials/types-and-consts.ts
olegnn:
```
Object.getOwnPropertyNames({}.__proto__) =>
(12) ['constructor', '__defineGetter__', '__defineSetter__', 'hasOwnProperty', '__lookupGetter__', '__lookupSetter__', 'isPrototypeOf', 'propertyIsEnumerable', 'toString', 'valueOf', '__proto__', 'toLocaleString']
```
When using an object as a map it's always better to nullify the prototype in order to avoid cases like `({})['constructor']` etc. While the current call might not be the case, any function which takes user input and then attempts to take a value from this map might be vulnerable.

src/anonymous-credentials/presentation.ts
olegnn: I'll move this logic to a single function as it will be changed anyway

src/anonymous-credentials/presentation-builder.ts
lovesh: The extra cost would be while making the first wasm call that will serialize and deserialize the params

src/anonymous-credentials/types-and-consts.ts
lovesh: This (unlikely) problem exists at so many places in this codebase :smile: . Anyway, thanks for explaining.

lovesh (approved):

PR #19, opened 2023-05-10
feat: pseudonyms supports in presentation

### LINKED ISSUE #18: Add pseudonyms support for presentation
Hi! 

I will be using dock libraries and dock wallet for https://github.com/privency

The [voting system](https://github.com/privency/poc/tree/main/vc-flow#poc-to-prove-how-to-use-anoncred-for-counting-votes) relies heavily on hash of credential attributes - aka pseudonyms.

Pseudonyms have been implemented in `Pseudonym.js`, and thoroughly tested in `pseudonym.spec.ts`, but they are not available in Presentations.

I worked on adding support for pseudonyms in presentation, allowing:
- pseudonyms based on secret only
- attributes-bound pseudonyms based on attributes only. Attributes can be taken from different credentials. 
- attributes-bound pseudonyms based on attributes + secret. Attributes can be taken from different credentials. 
- presenting the pseudonyms
- only presenting zk proof of pseudonym calculation, without revealing pseudonym itself. Possibility to also present a proof that the calculated pseudonym is equal to another unrevealed attribute

I am sending a draft PR, and would love your feedback.
After work is finished here, I'm happy to help bringing support for it in the wallet.

nicobao:
Closes #18 

Things I am not sure about and would particularly love your help with:
- does the statement calculated from `Statement.attributeBoundPseudonym` and sent in the proofSpec with the presentation already contain the pseudonym? If I understand well, it creates a Pedersen commitment but does not reveal the pseudonym? It is critical to me because in Privency protocol, it is sometimes necessary to keep pseudonyms private (full protocol description is here: https://github.com/privency/poc/tree/main/vc-flow#poc-to-prove-how-to-use-anoncred-for-counting-votes )
- how do I prove that the revealed pseudonym I have put somewhere in the presentation sperc is the one that correspond to the attached proof?
- where to put the pseudonym I want to reveal in the presentation spec? do I have to put it in the credentials themselves to be able to prove equality between the pseudonym and another attribute afterwards? But if the pseudonym is calculated from multiple credential attributes, in which credentials should I put it in? Can I provide this kind of equality proof if I don't present the pseudonym in a credential but outside, like with `attributeEqualities` (option I chose for now)?
- am I supposed to decode the pseudonym I want to reveal or should I keep it encoded? What about the scope? I suggested to use the scope as a dictionary key, so it must be a string or a number.

Known lacking elements (should be implemented before PR is ready)
- no support for plain pseudonym (no attribute-bound)
- secret is mandatory and it shouldn't be for attribute-bound pseudonyms
- no unit tests


lovesh:
Hi. Thanks for the PR.

> does the statement calculated from Statement.attributeBoundPseudonym and sent in the proofSpec with the presentation already contain the pseudonym?

Yes. But generally, the proofSpec (object of class `ProofSpecG1`) isn't meant to be sent with the presentation but should be created by the verifier.  So the prover sends the pseudonym but not the statement as it's expected that the verifier will know the commitment key independently of the prover.

> If I understand well, it creates a Pedersen commitment but does not reveal the pseudonym

The Pedersen commitment is the `pseudonym` as seen [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/src/Pseudonym.ts#L25). 

> it is sometimes necessary to keep pseudonyms private (full protocol description is here: https://github.com/privency/poc/tree/main/vc-flow#poc-to-prove-how-to-use-anoncred-for-counting-votes )

Can you tell me specifically where you need a pseudonym in the above diagram and why do you need them to be private?

> how do I prove that the revealed pseudonym I have put somewhere in the presentation sperc is the one that correspond to the attached proof

 The proof itself is proving that you know the secret key of the pseudonym. And pseudonyms are not yet integrated in the presentation spec but otherwise you can see [this test](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/pseudonyms.spec.ts). 

> where to put the pseudonym I want to reveal in the presentation spec? do I have to put it in the credentials themselves to be able to prove equality between the pseudonym and another attribute afterwards? But if the pseudonym is calculated from multiple credential attributes, in which credentials should I put it in? Can I provide this kind of equality proof if I don't present the pseudonym in a credential but outside, like with attributeEqualities (option I chose for now)?

Pseudonyms not bound to any attribute should definitely be outside of any credentials and be a top-level key like `attributeEqualities`. And as you mentioned for bound attributes, they might be bound to attributes across credentials so better to keep them at the top level as well.  

>  am I supposed to decode the pseudonym I want to reveal or should I keep it encoded? What about the scope? I suggested using the scope as a dictionary key, so it must be a string or a number.

The crypto logic needs bytearrays but while sending them in presentation, you should encode it to base64. Similarly for scope, it needs to be bytearray for crypto logic but in a presentation in can be kept as a string.

> secret is mandatory and it shouldn't be for attribute-bound pseudonyms

Have you seen [this note](https://github.com/docknetwork/crypto-wasm-ts/blob/master/src/Pseudonym.ts#L30)?

nicobao:
> Hi. Thanks for the PR.
> 
> > does the statement calculated from Statement.attributeBoundPseudonym and sent in the proofSpec with the presentation already contain the pseudonym?
> 
> Yes. But generally, the proofSpec (object of class `ProofSpecG1`) isn't meant to be sent with the presentation but should be created by the verifier. So the prover sends the pseudonym but not the statement as it's expected that the verifier will know the commitment key independently of the prover.
> 
Ok, so that mean the pseudonym cannot be kept private.
> > If I understand well, it creates a Pedersen commitment but does not reveal the pseudonym
> 
> The Pedersen commitment is the `pseudonym` as seen [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/src/Pseudonym.ts#L25).
> 
> > it is sometimes necessary to keep pseudonyms private (full protocol description is here: https://github.com/privency/poc/tree/main/vc-flow#poc-to-prove-how-to-use-anoncred-for-counting-votes )
> 
> Can you tell me specifically where you need a pseudonym in the above diagram and why do you need them to be private?
> 

At the first step, registration, users send a hash of their university ID (`hash_ID`k and a hash of a secret `hash_secret`) to Privency, that is kept in a database associated with the Oauth2 account.

Then Privency sends users a VC containing the Hash ID and the Hash Secret called the "registration VC".

Whenever users want to vote they need to provide :
- another hash calculated from the same secret that serves as unique private identifier. Two option here to know that it's the same secret. Either you prove you can recalculate the `hash secret`, without presenting the `hash secret`. Or the secret used to calculate the `hash_secret` was blind-signed together with other attributes in the `registration VC`. So we just expect a hash from the secret attribute.
- proof that users still hold a valid unrevoked university ID that still hash to the hash ID contained in the `registration VC`, without revealing the `hash ID`. Another option, like above, is to blind-sign the university ID in the `registration VC` ad simply prove equality between this one and the original university VC (which is an unrevoked credential).

> > how do I prove that the revealed pseudonym I have put somewhere in the presentation sperc is the one that correspond to the attached proof
> 
> The proof itself is proving that you know the secret key of the pseudonym. And pseudonyms are not yet integrated in the presentation spec but otherwise you can see [this test](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/pseudonyms.spec.ts).

Yes, I noticed, and it's exactly the purpose of this PR: integrate pseudonyms in the presentation spec. How do I prove the order of presentation? Imagine the verifier requests many pseudonyms, how do the prover prove which corresponds to what in the presentation?

> 
> > where to put the pseudonym I want to reveal in the presentation spec? do I have to put it in the credentials themselves to be able to prove equality between the pseudonym and another attribute afterwards? But if the pseudonym is calculated from multiple credential attributes, in which credentials should I put it in? Can I provide this kind of equality proof if I don't present the pseudonym in a credential but outside, like with attributeEqualities (option I chose for now)?
> 
> Pseudonyms not bound to any attribute should definitely be outside of any credentials and be a top-level key like `attributeEqualities`. And as you mentioned for bound attributes, they might be bound to attributes across credentials so better to keep them at the top level as well.
> 

Good, that's what I did!

> > am I supposed to decode the pseudonym I want to reveal or should I keep it encoded? What about the scope? I suggested using the scope as a dictionary key, so it must be a string or a number.
> 
> The crypto logic needs bytearrays but while sending them in presentation, you should encode it to base64. Similarly for scope, it needs to be bytearray for crypto logic but in a presentation in can be kept as a string.
> 
Ok good, that's what I did!

> > secret is mandatory and it shouldn't be for attribute-bound pseudonyms
> 
> Have you seen [this note](https://github.com/docknetwork/crypto-wasm-ts/blob/master/src/Pseudonym.ts#L30)?
Hmm yes but I think there are tests in `pseudonym.spec.ts` that do not require `secret` for attributeBoundPseudonym. What I mean is that my current implementation in presentation does not allow attributeBoundPseudonym to not have a secret, while it is implemented in the layer below.

nicobao:
> > Ok, so that mean the pseudonym cannot be kept private... Is it hard to change that?
> 
> Yes, support needs to be built in the Rust library for this, and since the pseudonym itself is a commitment, proving its knowledge will likely require pairings which are expensive so lets try to avoid them.

:+1: 

> 
> > Whenever users want to vote they need to provide :
> > another hash calculated from the same secret that serves as unique private identifier for the vote. Two option here to know that it's the same secret.
> > Either you prove you can recalculate the hash secret, without presenting the hash secret. Why keeping the hash_secret private from Privency? Because otherwise, we'd know exactly which Oauth2 account voted what, and we want to avoid that to preserve privacy.
> > Or the secret used to calculate the hash_secret was blind-signed together with other attributes in the registration VC. So we just expect a hash from the secret attribute.
> 
> Yes, let use bling signing. While requesting `registration VC`, the user blinds the secret (not its hash) so that the issuer is not aware of it but the VC still contains the secret as an attribute. Now during voting (presentation), the user shares an attribute-bound pseudonym where the attribute is the secret. Create it without the "secretKey" like [this](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/pseudonyms.spec.ts#L263).

:+1: 

How to prove that the secretKey that is blind-signed is the same that was used to create the hash?
Do Privency need to issue a transitory VC with the random blinded-secret, and then request from the prover a hash from the secret attribute in the VC Privency just issued? Or can this be avoided?

> 
> > proof that users still hold a valid unrevoked university ID that was the same ID as the one used during registration
> 
> This can be simply achieved by revoking `registration VC`. Shouldn't need blinding or any equality proof.
> 

Hmm no. There are two VC:
- the university VC issued by the university
- the registration VC issued by Privency

In the university VC, there is a unique university ID per university member.
When issuing the registration VC, Privency blind-signs the very same ID in registration VC.
Privency uses a hash of this university ID to bound it to a Oauth2 account. This is necessary to avoid university members to create plenty of accounts to vote multiple times.

Between Privency registration and voting, the university VC could be revoked, but Privency wouldn't know and wouldn't know what registration VC to revoke anyway, because Privency only knows a hash of the ID, and this is by design because we want to give as few information to Privency as possible. Privency should not alone be able to know who registered. The only attack vector for Privency to know who registered would be if Privency was colluding with the University and try all the hash of university ID to match them with Privency database. It's unfortunately impossible to make it more private, but we think this situation is acceptable. The most important is that no matter what, it is cryptographically impossible to know who voted what: this is achieved using the secret.

Now, when voting, Privency needs to verify that:
- the university VC is valid and has not been revoked (easy)
- the registration VC is valid and has not been revoked (easy)
**- the university ID in the registration VC is equal to the university ID in the university VC**

I need an equality check for my use-case, and blind-sign the university ID, not just the secret.

> > How do I prove the order of presentation? Imagine the verifier requests many pseudonyms, how do the prover prove which corresponds to what in the presentation?
> 
> Have 2 keys `boundedPseudonyms` and `unboundedPseudonyms` and each will be an array like below
> 
> ```
> unboundedPseudonyms: [
>     {commitKey: <a reference to a commitment key so that verifier can fetch it, we use `paramId`, pseudonym: <encoded value>},
>     {commitKey: <a ref.... },
> ],
> boundedPseudonyms: [
>     {
>         commitKey: <a ref....>,  
>         pseudonym: <encoded value>,
>         attributes: [
>             {credIdx: <index of credential in PresentationSpecification.credentials>, attribute: <full attribute name>},
>             {credIdx: <index of cred.. }
>         ]
>     },
>     {
>         commitKey: <a ref....>,  
>         pseudonym: <encoded value>,
>         attributes: [
>             {credIdx: <index of credential in PresentationSpecification.credentials>, attribute: <full attribute name>},
>             {credIdx: <index of cred.. }
>         ]
>     }
> ]
> ```
> 
> What do you think?

It looks good, I'm going to edit my PR to add this!

> 
> > Hmm yes but I think there are tests in pseudonym.spec.ts that do not require secret for attributeBoundPseudonym. What I mean is that my current implementation in the presentation does not allow attributeBoundPseudonym to not have a secret, while it is implemented in the layer below.
> 
> Some of them do but in your case its not needed.

Thank you again for your help :pray:

nicobao:
> > How to prove that the secretKey that is blind-signed is the same that was used to create the hash?
> 
> The "hash" should serve as a pseudonym, right? The don't hash but commit to the attribute (which the psedonym is). Adding witness equality like this https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/pseudonyms.spec.ts#L313 will ensure that above condition is met. It will generate an equality proof between the attribute the secret of the psedonym. From now on, i will assume whenever you say hash of attribute, you mean a an attribute bound pseudonym so please correct me when you really need a hash.

Yes, sorry for some reasons I keep associating the pseudonym with a hash, but it's a pedersen commitment, not a hash.

The thing I am wondering is how to send the blinded secretKey together with the proof that the secretKey is the same that was used to present the pseudonym? At this point, the secretKet is not yet part of a credential, but the witness equality you pointed seems to only work with attributes included in existing credentials. Am I mistaken?

> 
> > can this transitory VC be avoided?
> 
> Yes, no need of this.

:+1: 

> 
> > In the university VC, there is a unique university ID per university member.
> > When issuing the registration VC, Privency blind-signs the very same ID in registration VC.
> > Privency uses a hash of this university ID to bound it to a Oauth2/FIDO2 account. This is necessary to prevent university members from creating plenty of accounts to vote multiple times.
> 
> +1
> 
> > Between Privency registration and voting, the university VC could be revoked, but Privency wouldn't know and wouldn't know what registration VC to revoke anyway, because the university ID is a private information between the Holder and the University, and Privency only knows a hash of the ID
> 
> Whats the revocation mechanism of the university VC? Accumulators? If yes then during voting, the user could prove that his university VC isn't revoked without revealing his university id. If the revocation mechanism is through an expiry date, then the user could do a range proof on the expiry date.

Technically the university VC must be a BBS+ VC for the proofs to work.
As of the revocation, there is no requirement. I wonder what's the classic revocation mechanism in case issuers use did:web?
I am wondering if there is any way to present a BBS+ VC when the originally issued VC is another type of VC?

> 
> > the university ID in the registration VC is equal to the university ID in the university VC
> > I need an equality check for my use-case, and blind-sign the university ID, not just the secret.
> 
> This is doable easily, right?

Yes!

> 
> > For the attributes field, I'd go for the following:
> 
> ```
> unboundedPseudonyms: [
>     {commitKey: <a reference to a commitment key so that verifier can fetch it, we use `paramId`, pseudonym: <encoded value>},
>     {commitKey: <a ref.... },
> ],
> boundedPseudonyms: [
>     {
>         commitKey: <a ref....>,  
>         pseudonym: <encoded value>,
>         attributes: { // key is credential index
>             1: [<full attribute name1>, <full attribute name2>,...],
>             2: [<full attribute name3>, <full attribute name4>,...],
>         }
>     },
>     {
>         commitKey: <a ref....>,  
>         pseudonym: <encoded value>,
>         attributes: {
>             1: [<full attribute name1>, <full attribute name2>,...],
>             2: [<full attribute name3>, <full attribute name4>,...],
>         }
>     }
> ]
> ```
> 
> Thats better. Lets use this.

For the `boundedPseudonyms`, I finallly went for:
```
boundedPseudonyms: {
    // this unique key is calculated by doing:
    // basesForAttributes.map(b => b.toString()).join('')+basesForSecretKey.toString()
    // (@see source code)
    // I find this handy for picking the right pseudonym on the verifier side
    // verifier does not need to do "boundedPseudoyms.find(...)" to get the right pseudonym, which is O(n), 
    // they can just get the value directly from the map, which is O(1)
    // and if necessary it's still possible to iterate through the pseudonyms by doing: "boundedPseudonyms.entries()"
    "unique_key_1":
          {
              commitKey: <a ref....>,  
              pseudonym: <encoded value>,
              attributes: { // key is credential index
                  1: [<full attribute name1>, <full attribute name2>,...],
                  2: [<full attribute name3>, <full attribute name4>,...],
              }
          },
   "unique_key_2":
          {
              commitKey: <a ref....>,  
              pseudonym: <encoded value>,
              attributes: {
                  1: [<full attribute name1>, <full attribute name2>,...],
                  2: [<full attribute name3>, <full attribute name4>,...],
              }
          }
}
```
I'd do something similar with the `unboundedPseudonyms`.

> 
> > My understanding is that the commitKeys are baseforSecretKey (if defined) and basesForAttributes for boundedPseudonyms and simply base for unboundedPseudonyms. Am I correct?
> 
> Correct.

Good.

> 
> > How does the verifier verifies that the pseudonym he extracts from the spec above corresponds to the Statements and Witnesses sent by the prover in proofSpec? Is there something to implement in Presentation.verify function?
> 
> The provided proof proves that user knows secret/attribute (and the attribute is same in case of bound) behind psedonym. So in `Presentation.verify`, construct `Statement.pseudonym`s, `Statement.attributeBoundPseudonym` and `WitnessEqualityMetaStatement` accordingly.

Makes sense. I implemented what you suggested, but I still get an error in the unit test that check whether the verification goes well:
- https://github.com/nicobao/crypto-wasm-ts/blob/7d03de485323a0f2ba6a4bf5d3ec7c12c7f14532/tests/anonymous-credentials/presentation.spec.ts#L556
- https://github.com/nicobao/crypto-wasm-ts/blob/7d03de485323a0f2ba6a4bf5d3ec7c12c7f14532/src/anonymous-credentials/presentation.ts#L305

I don't understand what's wrong... I have put some comments in the code.
Could you help me?

I almost finished the PR for the boundedPseudonyms, then I'll do unboundedPseudonyms and add some more tests and it will be ready for full review.

lovesh:
> The thing I am wondering is how to send the blinded secretKey together with the proof that the secretKey is the same that was used to present the pseudonym? At this point, the secretKet is not yet part of a credential, but the witness equality you pointed seems to only work with attributes included in existing credentials.

This is regarding the request for `registration VC`? Then the `secretKey` is already part of the university VC. The flow i had in mind was-
1. User requests a `university VC` which contains the `secretKey` as an attribute but is blinded (hidden) from issuer.
2. User requests a `registration VC` by sharing a presentation from `university VC`, the pseudonym created from `secretKey` and proof that `secretKey` in pseudonym is equal to the one in  `university VC`

> As of the revocation, the only requirement is to be privacy-preserving. I wonder what's the classic revocation mechanism in case issuers use did:web?

The revocation mechanism used is orthogonal to the DID method.

>  I am wondering if there is any way to present a BBS+ VC when the university VC originally issued is another type of VC (AnonCred with CL signature or basic JWT/JSON-LD VC for example)?

The CL sigs I have seen use RSA and with different group size than the one we use with elliptic curves so proving attribute equality would be difficult and we don't support that even in the Rust library.

> For the boundedPseudonyms, I finallly went for:
   // this unique key is calculated by doing:
    // basesForAttributes.map(b => b.toString()).join('')+basesForSecretKey.toString()
    
This can have collisions for the key when the same bases are used to create many pseudonyms from different credentials. The pseudonym itself being the key is better as they are required to be unique (in a scope) and have a negligible chance of collision.

> I implemented what you suggested, but I still get an error in the unit test that check whether the verification goes well:

Will check. Can you please merge the latest master branch in your PR? Most of the new changes were part of this [PR]( https://github.com/docknetwork/crypto-wasm-ts/pull/17). These bring in support for additional schemes but should not impact your logic much.

nicobao:
> > The thing I am wondering is how to send the blinded secretKey together with the proof that the secretKey is the same that was used to present the pseudonym? At this point, the secretKet is not yet part of a credential, but the witness equality you pointed seems to only work with attributes included in existing credentials.
> 
> This is regarding the request for `registration VC`? Then the `secretKey` is already part of the university VC. The flow i had in mind was-
> 
>     1. User requests a `university VC` which contains the `secretKey` as an attribute but is blinded (hidden) from issuer.
> 
>     2. User requests a `registration VC` by sharing a presentation from `university VC`, the pseudonym created from `secretKey` and proof that `secretKey` in pseudonym is equal to the one in  `university VC`
> 

It cannot work for two reasons:
- we can't expect to control what's the university VC look like
- there should be 1 secret per poll, so that if you lose it you only lose the possibility to change your vote for the polls that you already registered on, but you can still vote in other polls. Also, if we were using 1 secret per university VC that means we'd need to expect the university to keep track of secret pseudonym and refuse issuing new VC to users in case they lost their secret... Otherwise their vote could be counted twice after they revoked and got a new VC with a new secret. It would be unrealistic to expect the university to manage such complexity.

There are two use-cases for our product:
1. we are the university VC issuer
1. we are not the university issuer - in that case we have no control over the VC initially issued by the university

At the beginning we will ship case 1/ - but the goal of the project is really 2/: we want to support any existing VCs - with as little limitations as possible.
Because the goal is 2/ I don't want to make a design for 1/ that does not work for 2/.

My idea would be the following:
- the only expectation from the university VC issuer is to issue an attribute that represents a unique identifier per holder in the VC. Preferentially a cryptographically random identifier.
- on user starting the registration process to a specific poll, we request the proof of ownership of a university VC, then two options:
    - if it's a BBS+ VC, then move on to the next step
    - if it's not, then blind all the attributes from the VC, send it to us and we issue a new BBS+ VC that's BBS+. It would be even better if the wallet itself could do that internally - sending the new BBS+ VC with a proof of ownership of the other one... Is that step possible? How so? I am not sure right now (see below).
- we request the user to send a blinded secret
- we send a `poll VC` to the user, that contains information about the poll, and the blinded secret (I was wondering whether this transitory step is necessary and it seems it is)
- then the user can request a `registration VC` and we proceed like discussed except the secret equality is associated with the `poll VC` just issued

That seems to work fine to me, so long as the university VC type is BBS+. Not so sure otherwise...

> 
> > As of the revocation, the only requirement is to be privacy-preserving. I wonder what's the classic revocation mechanism in case issuers use did:web?
> 
> The revocation mechanism used is orthogonal to the DID method.

As of the `poll VC` and `registration VC`, users need to trust us anyway, so I am thinking we could use a [status list](https://www.w3.org/TR/vc-status-list/). The list would be limited to the people who've already registered to the poll, so it should be manageable in terms of performance. We would present that list to the holder anytime they need to present a non-revocation proof. What's the state in the art here? Is there a better option?

As of the university VC, I am not sure which method is best yet.

> 
> > I am wondering if there is any way to present a BBS+ VC when the university VC originally issued is another type of VC (AnonCred with CL signature or basic JWT/JSON-LD VC for example)?
> 
> The CL sigs I have seen use RSA and with different group size than the one we use with elliptic curves so proving attribute equality would be difficult and we don't support that even in the Rust library.

Hmm... So, regarding of my suggestion above, that means:
- proving ownership of a non-BBS+ university VC should be doable
- blinding non-BBS+ university VC attributes is doable
- but proving equality between those blinded attributes and the ones in the non-BBS+ VC is not doable

Which would mean we cannot support non-BBS+ university VC - am I correct?

> 
> > For the boundedPseudonyms, I finallly went for:
> > // this unique key is calculated by doing:
> > // basesForAttributes.map(b => b.toString()).join('')+basesForSecretKey.toString()
> 
> This can have collisions for the key when the same bases are used to create many pseudonyms from different credentials. The pseudonym itself being the key is better as they are required to be unique (in a scope) and have a negligible chance of collision.
> 

Oh right, I didn't realize that! Changing this now.

> > I implemented what you suggested, but I still get an error in the unit test that check whether the verification goes well:
> 
> Will check. Can you please merge the latest master branch in your PR? Most of the new changes were part of this [PR](https://github.com/docknetwork/crypto-wasm-ts/pull/17). These bring in support for additional schemes but should not impact your logic much.

I just rebased my branch, it is up to date! Thanks for checking.

lovesh:
> It cannot work for two reasons:
    we can't expect to control what's the university VC look like
    there should be 1 secret per poll, so that if you lose it you only lose the possibility to change your vote for the polls that you already registered on, but you can still vote in other polls. Also, if we were using 1 secret per university VC that means we'd need to expect the university to keep track of secret pseudonym and refuse issuing new VC to users in case they lost their secret... Otherwise their vote could be counted twice after they revoked and got a new VC with a new secret. It would be unrealistic to expect the university to manage such complexity.

Sorry, i forgot about these details. So the flow should be this

1. User requests a `university VC` which contains the student id attribute called `sId` and no attribute is blinded.
2. User requests a `registration VC` by sharing a presentation from `university VC`. The user blinds 2 attributes, `sId` and `secretKey` and proves that blinded `sId` is equal to the `sId` attribute from `university VC`.
3. While participating in the poll, the user creates a pseudonym bounded to 2 attributes, `sId` and `secretKey` of above `registration VC`

Does the above work?

> if it's not, then blind all the attributes from the VC, send it to us and we issue a new BBS+ VC that's BBS+. It would be even better if the wallet itself could do that internally - sending the new BBS+ VC with a proof of ownership of the other one... Is that step possible? How so? I am not sure right now (see below).

If the university VC is not of a supported scheme like BBS+, PS, BBS, above won't be possible at the moment and would be a significant effort even if we restrict the signature scheme of university VC to EdDSA, ECDSA. We will need to use a zk-SNARK to prove that you know a EdDSA signature and the attributes and that blinding has been done correctly and all this will be proved inside the snark circuit making it quite expensive.

> we send a poll VC to the user, that contains information about the poll, and the blinded secret (I was wondering whether this transitory step is necessary and it seems it is)

Yes, you don't need it.

> As of the poll VC and registration VC, users need to trust us anyway, so I am thinking we could use a status list. The list would be limited to the people who've already registered to the poll, so it should be manageable in terms of performance.

The problem with the status list isn't performance but the lack of privacy because the verifier will learn `statusListIndex` and know which "registered user" it is. If that is acceptable then using status list is fine.

> Is there a better option?

Better in terms of?

> proving ownership of a non-BBS+ university VC should be doable
    blinding non-BBS+ university VC attributes should be doable
    but proving equality between those blinded attributes and the ones in the non-BBS+ VC is NOT doable

I can't answer the question without knowing which sig scheme is being used in university VC but it's quite expensive and not supported atm for EdDSA, ECDSA.

nicobao:
> > It cannot work for two reasons:
> > we can't expect to control what's the university VC look like
> > there should be 1 secret per poll, so that if you lose it you only lose the possibility to change your vote for the polls that you already registered on, but you can still vote in other polls. Also, if we were using 1 secret per university VC that means we'd need to expect the university to keep track of secret pseudonym and refuse issuing new VC to users in case they lost their secret... Otherwise their vote could be counted twice after they revoked and got a new VC with a new secret. It would be unrealistic to expect the university to manage such complexity.
> 
> Sorry, i forgot about these details. So the flow should be this
> 
>     1. User requests a `university VC` which contains the student id attribute called `sId` and no attribute is blinded.
> 
>     2. User requests a `registration VC` by sharing a presentation from `university VC`. The user blinds 2 attributes, `sId` and `secretKey` and proves that blinded `sId` is equal to the `sId` attribute from `university VC`.

That's right but before the registration VC is sent, we need to record certain information.

Our product needs to keep track of an anonymized version of `secretKey` and `sId` per poll and per user. A new OAuth2 account should not be allowed to bind the same `sId` to his account and be allowed to vote in the poll as the same university member again - unless the first account is unregistered. Similarly, after revocation and request of a new `registration VC`, the user should provide the same secret, otherwise he'd be able to vote twice.

My original idea for that was for our product to request two pseudonyms before sending the `registration VC`:
- one created from `secretKey` + a poll/user-specific nonce - let's call it `psecretKey`
- one created from `sId` + another poll/user-specific nonce - let's call it `psId`
- 
Nonce is useful to add randomness, especially for `sId`, that may not be as cryptographically secure as we'd want.
Then we'd keep these pseudonyms in a database bounded to the OAuth2 user. The nonces also have to be kept in database. This way, if only `psId` leaks but not the nonce, the University won't be able to link `psId` with `sId` and know who registered.

> 
>     3. While participating in the poll, the user creates a pseudonym bounded to 2 attributes, `sId` and `secretKey` of above `registration VC`

I need to add a third attribute to this bounded pseudonym: yet another poll-specific nonce.
Otherwise, the same pseudonym would be used for all the polls created for that university, and we'd want to avoid that.

> 
> 
> Does the above work?
> 
> > if it's not, then blind all the attributes from the VC, send it to us and we issue a new BBS+ VC that's BBS+. It would be even better if the wallet itself could do that internally - sending the new BBS+ VC with a proof of ownership of the other one... Is that step possible? How so? I am not sure right now (see below).
> 
> If the university VC is not of a supported scheme like BBS+, PS, BBS, above won't be possible at the moment and would be a significant effort even if we restrict the signature scheme of university VC to EdDSA, ECDSA. We will need to use a zk-SNARK to prove that you know a EdDSA signature and the attributes and that blinding has been done correctly and all this will be proved inside the snark circuit making it quite expensive.

Alright, I can live with that restriction. BBS+ is where the community is heading anyway.

> 
> > we send a poll VC to the user, that contains information about the poll, and the blinded secret (I was wondering whether this transitory step is necessary and it seems it is)
> 
> Yes, you don't need it.

Even with these nonces and `psecretKey`/`psId`? 

> 
> > As of the poll VC and registration VC, users need to trust us anyway, so I am thinking we could use a status list. The list would be limited to the people who've already registered to the poll, so it should be manageable in terms of performance.
> 
> The problem with the status list isn't performance but the lack of privacy because the verifier will learn `statusListIndex` and know which "registered user" it is. If that is acceptable then using status list is fine.

The "registered user" could be yet another ID that is unique to each `registration VC` and generated by Privency on registration. We can call the attribute `rId`. It would only be used for the purpose of revocation.

This value would never be used for voting or anything else and should be kept private by the holder. Yes `rId` is private, but so long as the holder keeps it private, and we don't publicly associate this `rid` with `psId`, nobody can ever know which `psId` registered to the votes except us. Privency already know which `psId` have registered anyway, and users need to trust us to keep this information private, so I think that's acceptable, considering we would still need to collude with the University to know people's real identity (the `sId`).

The most important for the product is that it should be cryptographically infeasible to know who voted what - so long as the secretKey is kept private by the voter, even if Privency and the University collude or if there is a data leak.

> 
> > Is there a better option?
> 
> Better in terms of?

My requirements for the revocation mechanism of `registration VC` are:
- private (to the level exposed above)
- cheap
- scalable and performant
- widely adopted

The non-requirements (at the moment):
- decentralization - Privency already acts as an issuer so there is a minimum level of trust needed. Also, the `registration VCs` that Privency issue are not as long-lasting as governments use-case.

I am wondering if there is another revocation mechanism I simply missed that's already widely adopted?
I am thinking of rolling my own implementation but if there is already something available, that would be better...
Do Dock wallet support status lists already?

> 
> > proving ownership of a non-BBS+ university VC should be doable
> > blinding non-BBS+ university VC attributes should be doable
> > but proving equality between those blinded attributes and the ones in the non-BBS+ VC is NOT doable
> 
> I can't answer the question without knowing which sig scheme is being used in university VC but it's quite expensive and not supported atm for EdDSA, ECDSA.
> 
> > I just rebased my branch, it is up to date! Thanks for checking.
> 
> Thanks

I thank you! Love all the support you're providing!
Btw, I will be at [DICE Europe](https://diceurope.org/) in Zurich next month, I would love to see you there!

nicobao:
I just rebased and changed the "key" in the boundedPseudonyms presented Map to the decoded pseudonym instead of (basesForAttributes+basesForSecretKey).

lovesh:
> My original idea for that was for our product to request two pseudonyms before sending the registration VC:
one created from secretKey + a poll/user-specific nonce - let's call it psecretKey
one created from sId + another poll/user-specific nonce - let's call it psId

There is one additional thing that you need when using the nonce. The same user can register multiple times by using different nonces so it needs to be proven that the `sId` used in each `psId` is different. The naive approach (and the only one i can think of now) is for the registration service to publish these `psId`s and during registration the user proves that his `sId` is not used in any `psId`s. The crypto logic would be proving the `sId` in his pseudonym `psId = G*sId + H*r` is different from each existing pseudonym like `G*sId1 + H*r1, G*sId2 + H*r2, ...`, where the `r`s are nonces and `G` and `H` are public bases for commitment. But this is not practical (beyond few hundred pseudonyms) as each inequality requires a separate proof. With this you won't need `registration VC` as you can directly use the above strategy for the poll. However we don't support any specific kind of inequlity proofs and you have to use the zk-SNARK and our Circom integration. 
But if you can live with this limitation that pseudonym `psId` can be brute forced then having 2 pseudonyms as above (without nonce) makes sense. 
However the best approach for now would be to introduce another credential thats issued to a user only once by Privency (simialar to your transitory VC). Before a user can request registration VC, it uses the university VC in a presentation revealing his `sId` to get a blinded VC where user chooses a random value with sufficient entropy, say 256-bits, called `masterSecretKey`, blinds it and request a `Privency VC` which has the `masterSecretKey` as an attribute. Now on each request of `registration VC`, user creates a pseudonym bound to `masterSecretKey` without any nonce. But even this has the limitation that a user has the same pseudonym during different registration phases and is correlated though you can't tie it to any `sId`.

> I need to add a third attribute to this bounded pseudonym: yet another poll-specific nonce.
Otherwise, if users reuse the same secretKey, the same pseudonym would be used for all the polls created for that university, and we'd want to avoid that.

This will cause the same problem as above. I think its reasonable to expect users to generate a different `secretKey` for each poll.

> The "registered user" could be yet another ID that is unique to each registration VC and generated by Privency on registration. We can call the attribute rId. It would only be used for the purpose of revocation.

Privency will know which "registered user" corresponds to which `masterSecretKey`. If thats acceptable then you can take this route.

> I am wondering if there is another revocation mechanism I simply missed that's already widely adopted?

The 2 already mentioned seem the most appropriate. There is [a transparent construction of accumulators](https://eprint.iacr.org/2022/756) but I haven't looked at it much yet.

nicobao:
> > My original idea for that was for our product to request two pseudonyms before sending the registration VC:
> > one created from secretKey + a poll/user-specific nonce - let's call it psecretKey
> > one created from sId + another poll/user-specific nonce - let's call it psId
> 
> There is one additional thing that you need when using the nonce. The same user can register multiple times by using different nonces so it needs to be proven that the `sId` used in each `psId` is different. The naive approach (and the only one i can think of now) is for the registration service to publish these `psId`s and during registration the user proves that his `sId` is not used in any `psId`s. The crypto logic would be proving the `sId` in his pseudonym `psId = G*sId + H*r` is different from each existing pseudonym like `G*sId1 + H*r1, G*sId2 + H*r2, ...`, where the `r`s are nonces and `G` and `H` are public bases for commitment. But this is not practical (beyond few hundred pseudonyms) as each inequality requires a separate proof. With this you won't need `registration VC` as you can directly use the above strategy for the poll. However we don't support any specific kind of inequlity proofs and you have to use the zk-SNARK and our Circom integration. But if you can live with this limitation that pseudonym `psId` can be brute forced then having 2 pseudonyms as above (without nonce) makes sense. However the best approach for now would be to introduce another credential thats issued to a user only once by Privency (simialar to your transitory VC). Before a user can request registration VC, it uses the university VC in a presentation revealing his `sId` to get a blinded VC where user chooses a random value with sufficient entropy, say 256-bits, called `masterSecretKey`, blinds it and request a `Privency VC` which has the `masterSecretKey` as an attribute. Now on each request of `registration VC`, user creates a pseudonym bound to `masterSecretKey` without any nonce. But even this has the limitation that a user has the same pseudonym during different registration phases and is correlated though you can't tie it to any `sId`.

Hmm true. Interesting approach. I'd want to look into that after the first product is out.
I think Privency could generate and enforce a nonce that's poll-specific, not user/transaction-specific. By sending it in a Verifiable Credential pre-registration.

Three nonces:
- nonce for `sId` registration to `psId`
- nonce for `secretKey` registration to `psecretKey`
- nonce for voting, only used during voting in the specific poll.

Advantages of doing so:
- Using the nonce for registering `sId` to a poll means generating two different `psId` for two different polls: that would help to prevent Privency from knowing that the same underlying `sId` associated with a unique `psId` registered to two different polls. The nonce is not made public but it's enough for one user to make it public to compromise the brute-force attack protection it offers.
- Using the nonce for secretKey leads to two different `psecretKey` for each poll, meaning like for `psId` that we can't correlate them. The nonce is not public but it's enough for one user to make it public to compromise the brute-force attack protection it offers. SecretKey are likely to be secured though, as it will be automatically generated so brute-force attacks are not an issue.
- Using the nonce for the vote is the most important as the voting data is public. Having a nonce for each vote results in each user having a unique pseudonym for each poll. For a third-party viewer, that means every poll have a brand new set of unique pseudonyms.

> 
> > I need to add a third attribute to this bounded pseudonym: yet another poll-specific nonce.
> > Otherwise, if users reuse the same secretKey, the same pseudonym would be used for all the polls created for that university, and we'd want to avoid that.
> 
> This will cause the same problem as above. I think its reasonable to expect users to generate a different `secretKey` for each poll.
> 
> > The "registered user" could be yet another ID that is unique to each registration VC and generated by Privency on registration. We can call the attribute rId. It would only be used for the purpose of revocation.
> 
> Privency will know which "registered user" corresponds to which `masterSecretKey`. If thats acceptable then you can take this route.
> 
> > I am wondering if there is another revocation mechanism I simply missed that's already widely adopted?
> 
> The 2 already mentioned seem the most appropriate. There is [a transparent construction of accumulators](https://eprint.iacr.org/2022/756) but I haven't looked at it much yet.

Can I use accumulators without a ledger? I am not very familiar how it works.

I am adding support for unboundedPseudonyms now.
I am still not sure why the test does not verify. I get `Verifying proof returned error BBSPlusError(FirstSchnorrVerificationFailed)`. I guess that's because of the encoding/decoding method I used for passing around pseudonyms and bases (`new TextDecoder()` and `new TextEncoder()`), but I am not sure what else to use.

tests/anonymous-credentials/presentation.spec.ts
```diff
@@ -504,6 +521,58 @@ describe(`${Scheme} Presentation creation and verification`, () => {
     expect(presJson).toEqual(recreatedPres.toJSON());
   });
 
+  it('from credential1 - presenting pseudonyms', () => {
+    const builder1 = new PresentationBuilder();
```
lovesh, line 525: The encoding is the problem. Add this code here as a test
```
const decodedBaseForSecretKey = PseudonymBases.decodeBaseForSecretKey(base1ForSecretKey);
    const encodedBaseForSecretKey = PseudonymBases.encodeBaseForSecretKey(decodedBaseForSecretKey);
expect(base1ForSecretKey).toEqual(encodedBaseForSecretKey);
```

lovesh (changes requested):

lovesh:
>  I think Privency could generate and enforce a nonce that's poll-specific, not user/transaction-specific. By sending it in a Verifiable Credential pre-registration.

With this pre-registration VC, Privency would which student is willing to participate in the poll.

> Three nonces:

Like i mentioned before, if you are using the inequality proof, you can have just one poll specific pseudonym and thus only 1 nonce.

> Using the nonce for registering sId to a poll means generating two different psId for two different polls: that would help to prevent Privency from knowing that the same underlying sId associated with a unique psId registered to two different polls. Of course brute-force attacks are still possible if sId is not cryptographically secure, as Privency knows the nonce.

I have been meaning that Privency does not know the nonce. Because knowing the nonce makes brute forcing possible, eg `psId = G*sId + H*nonce` now Privency can compute `G*sId` as  `psId - H*nonce` and brute-force for `sId`

> Though I admit, using these nonces is not critical for the first version of the product, especially if it is assumed that users use a different secretKey for every poll, which we can make the default.

So for the first version, having the pseudonym `psId` without a nonce is acceptable?

> Can I use accumulators without a ledger? I am not very familiar how it works.

Yes. Infact this library does not assume a ledger. You can read more about accumulators here https://github.com/docknetwork/crypto-wasm#accumulator and here https://github.com/docknetwork/crypto-wasm-ts#accumulators.  

> I am still not sure why the test does not pass 

Added comments.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -492,6 +560,96 @@ export class PresentationBuilder extends Versioned {
...
+      const witnessEqs: [WitnessEqualityMetaStatement, number][] = [];
+      const attributeBoundPseudonym = this.boundedPseudonyms[i];
+      const attributes: Uint8Array[] = [];
+      for (const [credIdx, attributeNames] of attributeBoundPseudonym.attributeNames.entries()) {
+        // lots of the following is extracted from getRevealedAndUnrevealed and should probably be packaged in a function
+        const cred = this.credentials[credIdx][0];
+        const schema = cred.schema as CredentialSchema;
+        const messages = cred.serializeForSigning();
+        const encoder = schema.encoder;
+        const [names, encodedValues] = encoder.encodeMessageObject(messages);
```
lovesh, line 578: You don't need to encode again as its already been done. Cache the encoded values like [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/src/anonymous-credentials/presentation-builder.ts#L392)

lovesh (changes requested):

nicobao:
> > I think Privency could generate and enforce a nonce that's poll-specific, not user/transaction-specific. By sending it in a Verifiable Credential pre-registration.
> 
> With this pre-registration VC, Privency would which student is willing to participate in the poll.

I am confused about what you mean? With or without nonces (with my own version of "nonce known by Privency" - not yours), Privency will know which `psId` registred to the poll.

> 
> > Three nonces:
> 
> Like i mentioned before, if you are using the inequality proof, you can have just one poll specific pseudonym and thus only 1 nonce.

If the inequality proof solution can achieve good performance, it is a superior solution to the one I proposed. But I can't afford diving into Circom right now. (I am very fortunate you guys implemented the Pseudonym functionality already :+1: .) That's a great candidate for v2.

> 
> > Using the nonce for registering sId to a poll means generating two different psId for two different polls: that would help to prevent Privency from knowing that the same underlying sId associated with a unique psId registered to two different polls. Of course brute-force attacks are still possible if sId is not cryptographically secure, as Privency knows the nonce.
> 
> I have been meaning that Privency does not know the nonce. Because knowing the nonce makes brute forcing possible, eg `psId = G*sId + H*nonce` now Privency can compute `G*sId` as `psId - H*nonce` and brute-force for `sId`

Yes, you're right, let's forget about the nonce for generating `psId`. And all the other nonces actually. I'll just make sure the secretKey registered by voters is different for each poll.

> 
> > Though I admit, using these nonces is not critical for the first version of the product, especially if it is assumed that users use a different secretKey for every poll, which we can make the default.
> 
> So for the first version, having the pseudonym `psId` without a nonce is acceptable?

Definitely yes. In the first version of the product, I will have a personal relationship with "the University" and I may be able to convince them to issue a cryptographically secure `sId` for each student.
And even if I don't, the product would still offer a good security model, considering nobody would now who voted what.

Also, there is another functionality where we handle the issuing of credentials for the University (or any group). In that case, I will have total freedom to issue a cryptographically secure `sId` in the VCs.

> 
> > Can I use accumulators without a ledger? I am not very familiar how it works.
> 
> Yes. Infact this library does not assume a ledger. You can read more about accumulators [here](https://github.com/docknetwork/crypto-wasm#accumulator) and [here](https://github.com/docknetwork/crypto-wasm-ts#accumulators).

Nice! I will look into that.

> 
> > I am still not sure why the test does not pass
> 
> Added comments. Please don't force push for now. We can squash the commits later.

Thank you! Looking at it now! OK, I won't force push.

nicobao:
Here is the summary of how I would do the flow after our discussions.
Maybe I misunderstood some of your remarks, happy to get feedback.

Registration to a specific poll:
1. User logs in to Privency via OAuth2/WebAuthn
1. User starts the registration process for a specific poll
1. Before Privency issues anything, Privency needs to check if `sId` has never been registered or has been registered by the same user.
   1. the user sends `psId` from the University VC he holds. Privency keeps record of `psId`.
   1. Two possibilities:
	   1.If it's a first-time registration, Privency checks if `psId` is already used by another account. If yes, then stop, else continue
	   1.If the user had already registered (like from another device/wallet), Privency checks if the presented `psId` is different from the `psId` already associated with the user. If yes, then stop, else continue.
1. User blinds `sId` and sends it to Privency. Privency temporarilly records this blinded `sId` to issue `sId VC` and `registration VC`.
1. Privency issues a `sId VC` to the user containing the `sId` and the poll ID as attributes
1. User provides `psId` calculated from the `sId` attribute in `sId VC`
1. If `psId` does not match with the one previously recorded, stop there, else continue. This step is necessary to make sure the blinded `sId` the user sent to Privency is the same as the one in the University VC he holds.
1. User sends blinded `secretKey1` and blinded `secretKey2` to Privency. The secretKeys are cryptographically random value that's automatically generated on the wallet side for every poll. Privency temporarilly records these blinded secretKey (for the purpose of issuing `secret VC` and `registration VC`).
1. Privency issues a `secret VC` containing as attribute the poll ID, `secretKey1` and `secretKey2`.
1. User presents `psecretKey1`, a pseudonym bounded to attribute `secretKey1` attribute of `secret VC`. Privency records this pseudonym. Same for `psecretKey2`, pseudonym bounded to `secretKey2`.
1. Privency checks the validity of the secretKeys:
	1. If it's a first-time registration, do nothing and continue. 
	1. Else, Privency checks whether `psecretKey1` is different from the one that was already associated with the user or `psecretKey2 is different from the one that was already associated with the user. If there is a difference, stop, else continue.
1. Privency issues a `registration VC` that contains `secretKey1`, `secretKey2`, and the `sId` (using the blinded versions previously recorded), as well as the poll ID as attributes.
1. The `registration VC` will be used for voting in this specific poll (identified via the poll ID).

User is voting:
1. user does not need to be logged in
1. user provide non-revocation proofs for holding `registration VC` and `university VC`
1. user provides proof of equality that `sId` in his `registration VC` correponds to the one in his `university VC`
1. user sends a Pseudonym called `pvote` bounded to attributes `secretKey1`, `secretKey2`, and `sId` from the `registration VC` (`pvote=Gv*sId + Hv*secretKey1 + Hv'*secretKey2`). This pseudonym will be used as unique identifier for the vote. Explanation:
    - Even though Privency knows `psId`, `psecretKey1` and `psecretKey2` such that `psId=Gs*sId`, `psecretKey1=Gk*secretKey1` and `psecretKey2=Gk'*secretKey2`, `pvote` uses all three `sId`, `secretKey1` and `secretKey2`, and those values are kept secret from Privency. 
    - To be able to correlate `pvote` with `psId`, `psecretKey1` or `psecretKey2`, Privency would need to brute-force two of the hidden values like `secretKey1` and `sId` or `secretKey2` and `sId`. If Privency can guess `sId` and `secretKey1`, they could correlate `pvote` with `sId` by calculating `sId=(pvote - Gv*sId - Hv*secretKey1)/Gs`. 
    - However, even though brute-forcing `sId` from `psId` may be feasible because `sId` may not cryptographically secure, brute-forcing `secretKey1` or `secretKey2` is cryptographically impossible because they are cryptographically secure. 
    - Thefore, Privency cannot correlate `pvote` with the OAuth2 account bound to `psId` and so long as `secretKey1` and `secretKey2` are kept private by the voter, Privency cannot know who voted what. The reason why we need two secretKeys is because `sId` may not be cryptographically secure.


Note: 
- we could possibly ensure the secretKeys are secure enough on the verifier side, like by enforcing a proof when providing the blinded values.
- I omitted the classic VC-validity and non-revocation proof checks for the sake of simplicity
- on issuing a `registration VC` to the same user for the second time, we need to make sure the user uses the same secretKeys, otherwise he would end up having a different Pseudonym for the vote, and therefore be able to vote twice. By enforcing the same secretKeys, we make sure that when he votes again, his previous vote will simply be modified.
- on issuing a `registration VC` to a user, we make sure the `sId` is not already in use, because otherwise the same person identified by the `sId` could create multiple OAUth2 accounts to vote multiple times.

lovesh:
If the `university VC` has `sId`, I  don't see the need for `sId VC`. User can send a pseudonym bound to attribute `sId` while presenting `university VC` with the scope being the `poll Id`


> 7. If psId does not match with the one previously recorded, stop there, else continue. This step is necessary to make sure the blinded sId the user sent to Privency is the same as the one in the University VC he holds.

Does this step happen during the request of `registration VC` or even before that?

> 9. Privency issues a secret VC containing as attribute the poll ID, secretKey and secretKeyCommitKey.

`secret VC` is not needed as secret will be present in `registration VC`.

> 10. User presents psecretKey, a pseudonym bounded to secret VC attribute secretKey using attribute secretKeyCommitKey as base. Privency records this pseudonym.

This isn't needed because a pseudonym will be submitted later during voting anyway.

> 12. Privency issues a registration VC that contains secretKey and sId (using the blinded versions previously recorded), as well as secretKeyCommitKey, sIdCommitKey and the poll ID as attributes.

`secretKeyCommitKey`, sIdCommitKey` don't need to be attributes. You can have a convention that they are derived from `pollId` 

> 4. Privency sends off-band a baseForSecretKey and basesForAttributes. 
User verifies that baseForSecretKey!=secretKeyCommitKey and basesForAttributes!=sIdCommitKey

Privency can avoid sending these bases and thus the check as well.

> on issuing a registration VC to the same user for the second time, we need to make sure the user uses the same secretKey, otherwise he would end up having a different Pseudonym for the vote, and therefore be able to vote twice. 

Why would it be issued the second time for the same poll?


Have added a test [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts) for this use-case. There are comments in the test to explain my thinking. `signed1` is for `university VC` and `signed2` is for `registration VC`. `user-id` is what you call `sId` (sorry for the naming, was using existing test data)

src/Pseudonym.ts
```diff
@@ -64,6 +64,9 @@ export class AttributeBoundPseudonym extends BytearrayWrapper {
  * Used to create commitment key for pseudonyms
  */
 export class PseudonymBases {
+  static decoder = new TextDecoder();
```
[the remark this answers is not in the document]
nicobao: I used the `hex` representation you created for pseudonyms.

tests/anonymous-credentials/presentation.spec.ts
nicobao: Done

src/anonymous-credentials/presentation-builder.ts
nicobao: I removed it entirely, like you explained here: https://github.com/docknetwork/crypto-wasm-ts/pull/19#discussion_r1206692244

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -492,6 +560,96 @@ export class PresentationBuilder extends Versioned {
...
+      );
+      const baseForSecretKey = attributeBoundPseudonym.baseForSecretKey;
+      const decodedBaseForSecretKey =
+        baseForSecretKey !== undefined ? PseudonymBases.decodeBaseForSecretKey(baseForSecretKey) : undefined;
+      const decodedBasesForAttributes = PseudonymBases.decodeBasesForAttributes(
+        attributeBoundPseudonym.basesForAttributes
+      );
+      const decodedPseudonym = PseudonymBases.decoder.decode(pseudonym.value);
+      presentedBoundedPseudonyms.set(decodedPseudonym, {
+        commitKey: {
```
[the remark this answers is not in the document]
nicobao: I don't know what a reference to the params mean. Can you elaborate? What is it? How do you link the reference with the bases?
Do you expect verifier to keep track of the bases they sent to the prover to do the following?:
https://github.com/nicobao/crypto-wasm-ts/blob/6f3fb78322f2fba4bd6d6f1cfbc957cb93f0b00e/src/anonymous-credentials/presentation.ts#L296-L300

nicobao:
I just pushed the changes you requested, except for the reference to the bases (responded to your comments).
The tests pass now!

Thank you for doing the unit test for my use-case!

> If the `university VC` has `sId`, I don't see the need for `sId VC`. User can send a pseudonym bound to attribute `sId` while presenting `university VC` with the scope being the `poll Id`

The need is to make sure the `sId` that will be put in `registration VC` was actually taken from `university VC`.
The `sId VC` is just a way I found to work-around that problem. I don't know how to sending a blinded `sId` together with a proof that it was taken from `university VC`, without having this intermediary `sId VC`.

> 
> > 7. If psId does not match with the one previously recorded, stop there, else continue. This step is necessary to make sure the blinded sId the user sent to Privency is the same as the one in the University VC he holds.
> 
> Does this step happen during the request of `registration VC` or even before that?

During the same request flow. This step happens before `registration VC` is issued, to verify eligibility.

> 
> > 9. Privency issues a secret VC containing as attribute the poll ID, secretKey and secretKeyCommitKey.
> 
> `secret VC` is not needed as secret will be present in `registration VC`.

`secret VC` is a work-around to the following problem:
- i want to send a blinded secret together with a Pseudonym and a proof that this Pseudonym was calculated from this blinded secret.

I don't know how to do that without the `secret VC` intermediary. 

> 
> > 10. User presents psecretKey, a pseudonym bounded to secret VC attribute secretKey using attribute secretKeyCommitKey as base. Privency records this pseudonym.
> 
> This isn't needed because a pseudonym will be submitted later during voting anyway.

The pseudonym submitted during voting is not the same as the one that is recorded.
The pseudonym that is recorded is there to make sure users will always use the same secretKey.
For example, let's say you registered with a secretKey on your laptop, and then you use your phone to register again. We need to make sure the same secretKey is used, otherwise they could vote twice.

The Pseudonym that is used for voting should be unlinkable that the Pseudonym that was recorded during registration:
- the pseudonyms that are recorded during registration are:
    - an attribute-based pseudonym based on `sId`. The `sId` attribute will end up in `registration VC`.
    - a pseudonym based on a `secretKey`. The `secretKey` attribute will end up in `registration VC`.
- the Pseudonym that is used for voting is an attribute-based voting that uses `secretKey` and `sId` in `registration VC` and *different bases*. *Please can you confirm that using different bases make this pseudonym unlikable with registration pseudonyms described above?*

> 
> > 12. Privency issues a registration VC that contains secretKey and sId (using the blinded versions previously recorded), as well as secretKeyCommitKey, sIdCommitKey and the poll ID as attributes.
> 
> `secretKeyCommitKey`, sIdCommitKey`don't need to be attributes. You can have a convention that they are derived from`pollId`

Good point. Having different bases is key to ensure privacy though. Somehow the frontend should verify that at all times.

> 
> > 4. Privency sends off-band a baseForSecretKey and basesForAttributes.
> >    User verifies that baseForSecretKey!=secretKeyCommitKey and basesForAttributes!=sIdCommitKey
> 
> Privency can avoid sending these bases and thus the check as well.

Yea, I see.

> 
> > on issuing a registration VC to the same user for the second time, we need to make sure the user uses the same secretKey, otherwise he would end up having a different Pseudonym for the vote, and therefore be able to vote twice.
> 
> Why would it be issued the second time for the same poll?

Two different devices. Or the user lost the `registration VC`, and requests a new one. It must be the same `secretKey` otherwise if he votes again to the same poll, it would count twice (Privency have no way no know if people voted after registration, and Privency allows for users to change their vote over time.)

> 
> Have added a test [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts) for this use-case. There are comments in the test to explain my thinking. `signed1` is for `university VC` and `signed2` is for `registration VC`. `user-id` is what you call `sId` (sorry for the naming, was using existing test data)

lovesh:
> The need is to make sure the sId that will be put in registration VC was actually taken from university VC.
The sId VC is just a way I found to work-around that problem. I don't know how to send a blinded sId together with a proof that it was taken from university VC, without having this intermediary sId VC.

The test i shared does that. Its the equality enforced [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts#L150).

> i want to send a blinded secret together with a Pseudonym and a proof that this Pseudonym was calculated from this blinded secret.

See [this equality](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts#L263). I am creating the pseudonym from both `sId` and `secret` though but its the same idea.

> For example, let's say you registered with a secretKey on your phone, and then you use your laptop to register again. Or you lost your registration VC, so you request its revocation, and then you request a new one. We need to make sure the same secretKey is used, otherwise they could vote twice.

To get a registration VC, user has to submit a pseudonym to its `sId` and Privency can check if this pseudonym has requested a VC before or not. [This](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts#L76) is the one i am referring to. And the base here can be poll specific to ensure that requests across polls cant be correlated. In case of revocation, the VC won't be valid so any action using the VC should be rejected. So no compulsion to use the secret key. I didn't show revocation in test but you could use attribute `registration-id` for it.

> The Pseudonym that is used for voting should be unlinkable with the Pseudonym that was recorded during registration

They are unlinkable (in the test). 

> a pseudonym based on a secretKey. The secretKey attribute will end up in registration VC.

This shouldn't be needed.

> the Pseudonym that is used for voting is an attribute-based voting that uses secretKey and sId in registration VC and different bases. Please can you confirm that using different bases make this pseudonym unlikable with registration pseudonyms described above?

Yes.

nicobao:
> > The need is to make sure the sId that will be put in registration VC was actually taken from university VC.
> > The sId VC is just a way I found to work-around that problem. I don't know how to send a blinded sId together with a proof that it was taken from university VC, without having this intermediary sId VC.
> 
> The test i shared does that. Its the equality enforced [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts#L150).

:+1: 

> 
> > i want to send a blinded secret together with a Pseudonym and a proof that this Pseudonym was calculated from this blinded secret.
> 
> See [this equality](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts#L263). I am creating the pseudonym from both `sId` and `secret` though but its the same idea.

:+1: 

> 
> > For example, let's say you registered with a secretKey on your phone, and then you use your laptop to register again. Or you lost your registration VC, so you request its revocation, and then you request a new one. We need to make sure the same secretKey is used, otherwise they could vote twice.
> 
> To get a registration VC, user has to submit a pseudonym to its `sId` and Privency can check if this pseudonym has requested a VC before or not. [This](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/blind-sig-with-attribute-equality-and-pseudonym.spec.ts#L76) is the one i am referring to. And the base here can be poll specific to ensure that requests across polls cant be correlated. In case of revocation, the VC won't be valid so any action using the VC should be rejected. So no compulsion to use the secret key. I didn't show revocation in test but you could use attribute `registration-id` for it.
> 

"`In case of revocation, the VC won't be valid so any action using the VC should be rejected`"
=> yes but what if the action has _already_ taken place. Say someone voted before, with their valid `registration VC` when it was valid, before it was revoked. They used the secret for the vote, through a poll-specific unique identifier.

Say they later revoke the `registration VC`. OK, the new owner cannot vote with it again. And the old vote can still count, it is fine.
And then I issue a new `registration VC`, with a different secret, to the same user.
On attempting to vote to the same poll he already voted on, he will generate a new Pseudonym, that will act as a new identifier for the poll, identifier which is totally different and unlikable with the identifier the user used originally with the now-revoked `registration VC`. I will therefore have to count the vote twice, instead of modifying the old vote.

That is why, Privency must re-issue the same secret in the `registration VC`.

> > The Pseudonym that is used for voting should be unlinkable with the Pseudonym that was recorded during registration
> 
> They are unlinkable (in the test).

:+1: 

> 
> > a pseudonym based on a secretKey. The secretKey attribute will end up in registration VC.
> 
> This shouldn't be needed.
> 
> > the Pseudonym that is used for voting is an attribute-based voting that uses secretKey and sId in registration VC and different bases. Please can you confirm that using different bases make this pseudonym unlikable with registration pseudonyms described above?
> 
> Yes.

:+1: 

Thanks for the explanation on witness equality, it helps.

For your remark [comment](https://github.com/docknetwork/crypto-wasm-ts/pull/19#discussion_r1206704218), how do you want to proceed? Apart from that, I think the PR is ready on my side.

src/anonymous-credentials/presentation-builder.ts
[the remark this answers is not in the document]
lovesh: By reference, I meant a unique (in the pres. context) id of that param. Eg, The snark public key is called `pkId` in the pres. spec [here](https://github.com/nicobao/crypto-wasm-ts/blob/master/tests/anonymous-credentials/presentation.spec.ts#L961) and the verifier uses it like [this](https://github.com/nicobao/crypto-wasm-ts/blob/master/tests/anonymous-credentials/presentation.spec.ts#L968). This was to avoid creating multiple copies of a param object if the same param is used in multiple statements. But now i think what you have is fine because the bases wont be reused in multiple statements and these are not as big as snark keys so its fine.

src/Pseudonym.ts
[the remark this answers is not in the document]
lovesh: Can you please use base64 as that is used elsewhere as well.

lovesh:
>  And then I issue a new registration VC, with a different secret, to the same user.

Privency (issuer) will know if the same user is requesting a `registration VC` who was given earlier and which was revoked so it can deny the request. Is that behavior not acceptable?   

> For your https://github.com/docknetwork/crypto-wasm-ts/pull/19#discussion_r1206704218, how do you want to proceed? Apart from that, I think the PR is ready on my side.

Sorry about that, missed 2 comments. Replied now. 

Dock wallet does not support pseudonyms or most predicates in this repo yet.

nicobao:
> > And then I issue a new registration VC, with a different secret, to the same user.
> 
> Privency (issuer) will know if the same user is requesting a `registration VC` who was given earlier and which was revoked so it can deny the request. Is that behavior not acceptable?
> 

Not really, because this behavior would mean that once revoked, the user cannot never request a new `registration VC` again, and hence never vote to the poll (again - or not, we cannot know).
By enforcing the same secret on issuing a new `registration VC` to the same user, the user can still vote on the poll, so long as he remembers the secret.

But I think with your tremendous help and the unit test you wrote, I can implement this functionality now!
Thanks a lot!


> > For your [#19 (comment)](https://github.com/docknetwork/crypto-wasm-ts/pull/19#discussion_r1206704218), how do you want to proceed? Apart from that, I think the PR is ready on my side.
> 
> Sorry about that, missed 2 comments. Replied now.

Looking at it now.

> 
> Dock wallet does not support pseudonyms or most predicates in this repo yet.

I am choosing between:
- helping you add support for it in Dock Wallet => is it possible, and where would I need to start?
- integrating my own wallet within my PWA, using https://github.com/uport-project/veramo and `crypto-wasm-ts`

nicobao:
Just changed from `hex` to `base64` as you suggested.

Btw I noticed some linting issue on files unrelated to my changes when running `yarn pretty`:
```
	modified:   src/anonymous-credentials/credential-builder.ts
	modified:   src/anonymous-credentials/types-and-consts.ts
	modified:   src/anonymous-credentials/util.ts
	modified:   src/bbs-plus/params.ts
	modified:   src/index.ts
	modified:   src/ps/keys.ts
	modified:   src/ps/params.ts
	modified:   src/ps/signature.ts
	modified:   src/types.ts
```
Should I push it here?

src/anonymous-credentials/presentation-specification.ts
```diff
@@ -64,22 +64,49 @@ export interface IPresentedCredential {
...
+}
+
 /**
  * Specifies what the presentation is proving like what credentials, what's being revealed, which attributes are being proven
  * equal, bounds being enforced, etc
  */
 export class PresentationSpecification {
   credentials: IPresentedCredential[];
   attributeEqualities: AttributeEquality[];
+  // key: concatenated basesForAttributes+baseForSecretKey, encoded in UTF-8
```
lovesh, line 93: Outdated comment.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -492,6 +560,94 @@ export class PresentationBuilder extends Versioned {
...
+    }
+    this.spec.boundedPseudonyms = presentedBoundedPseudonyms;
+
+    // Create statements and witnesses for each unboundedPseudonyms
+    const presentedUnboundedPseudonyms: Map<string, IPresentedUnboundedPseudonym> = new Map();
+    for (let i = 0; i < this.unboundedPseudonyms.length; i++) {
+      const unboundedPseudonym = this.unboundedPseudonyms[i];
+      const pseudonym = Pseudonym.new(unboundedPseudonym.baseForSecretKey, unboundedPseudonym.secretKey);
+      const decodedBaseForSecretKey = PseudonymBases.decodeBaseForSecretKey(unboundedPseudonym.baseForSecretKey);
+      const decodedPseudonym = PseudonymBases.decode(pseudonym.value);
```
lovesh, line 638: This is slightly misleading as `pseudonym` is not a base. Better to define `encode`, `decode` on  `Pseudonym`. Use them above as well instead of `pseudonym.base64`

lovesh (changes requested):

lovesh:
> I can implement this functionality now

I will submit support for blind issuance in the anoncreds abstraction soon so you won't have to rely on lower level objects. 

> helping you add support for it in Dock Wallet => is it possible, and where would I need to start?

The wallet is closed source but I have asked if they can grant you access. 

> Btw I noticed some linting issue on files unrelated to my changes ...

Sure. Please merge the latest master as well.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -492,6 +560,94 @@ export class PresentationBuilder extends Versioned {
...
+        const encoder = schema.encoder;
+        const [names, encodedValues] = encoder.encodeMessageObject(messages);
+        let found = 0;
+        for (let j = 0; j < names.length; j++) {
+          if (attributeNames.has(names[j])) {
+            attributes.push(encodedValues[j]);
+            const witnessEq = new WitnessEqualityMetaStatement();
+            witnessEq.addWitnessRef(credIdx, j);
+            witnessEqs.push([witnessEq, attributes.length - 1]);
+            found++;
```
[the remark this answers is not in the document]
nicobao: I removed `found` entirely because of your previous comment

src/anonymous-credentials/presentation-builder.ts
nicobao: I find it weird to have `Pseudonym.encode(value)` and the `decode` equivalent when both `AttributeBoundPseudonym` and `Pseudonym` derive from `BytearrayWrapper`. The `encode` and `decode` static functions I use for AttributeBoundPseudonym are taken from Pseudonym in a very arbitrary way.
I'd find it more appropriate to use (static OR not) functions from `BytearrayWrapper`.
Though I think it is just a question of style, I'll do whatever you prefer.
I just implemented it in `Pseudonym` like you mentioned.

nicobao:
> > I can implement this functionality now
> 
> I will submit support for blind issuance in the anoncreds abstraction soon so you won't have to rely on lower level objects.
> 

:+1: 

> > helping you add support for it in Dock Wallet => is it possible, and where would I need to start?
> 
> The wallet is closed source but I have asked if they can grant you access.

Would be cool!

> 
> > Btw I noticed some linting issue on files unrelated to my changes ...
> 
> Sure. Please merge the latest master as well.

I just pushed the changes you requested, including the linting.
I rebased and fixup so you should `git pull --rebase` the branch instead of standard `git pull` if you want to push in it, to avoid issues.

PR #21, opened 2023-05-22
Fix error message + use strict equality checks + `eslint --fix`

olegnn:


src/anonymous-credentials/presentation.ts
```diff
@@ -361,7 +361,7 @@ export class Presentation extends Versioned {
         attributeEqualities: this.spec.attributeEqualities
       },
       attributeCiphertexts,
-      proof: b58.encode((this.proof as CompositeProofG1).bytes)
+      proof: b58.encode((this.proof).bytes)
```
lovesh, line 364: `(this.proof).bytes` -> `this.proof.bytes`

lovesh (approved):

PR #22, opened 2023-05-31
Improvements for error messages, comments and tests, minor fix

olegnn:


tests/scheme.ts
```diff
@@ -87,7 +88,7 @@ switch (Scheme) {
     CredentialBuilder = BBSCredentialBuilder;
     Credential = BBSCredential;
     encodeMessageForSigningIfPS = (msg) => msg;
```
lovesh, line 90: You don't have to define this as it same as previously defined.

tests/scheme.ts
```diff
@@ -110,7 +111,7 @@ switch (Scheme) {
     Credential = BBSPlusCredential;
     SignatureLabelBytes = BBS_PLUS_SIGNATURE_PARAMS_LABEL_BYTES;
     encodeMessageForSigningIfPS = (msg) => msg;
```
lovesh, line 113: Similar comment as above.

tests/scheme.ts
```diff
@@ -132,8 +133,9 @@ switch (Scheme) {
     Credential = PSCredential;
     SignatureLabelBytes = PS_SIGNATURE_PARAMS_LABEL_BYTES;
     encodeMessageForSigningIfPS = encodeMessageForSigning;
-    encodeMessageIfNotPS = (msg) => msg;
+    encodeMessageForSigningIfNotPS = (msg) => msg;
```
lovesh, line 136: You don't have to define this as it same as previously defined.

lovesh (changes requested):

src/ps/proof.ts
```diff
@@ -47,7 +47,7 @@ export class PSPoKSignatureProtocol {
         }
 
         const blinding = blindings.get(idx);
-        if (blinding !== void 0) {
+        if (blinding != null) {
```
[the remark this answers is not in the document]
olegnn: The idea is to align behavior with other functions as all of their checks are for `== null` (at least in `/ps` dir and the code I introduced in #17).
Although, I searched over the code, and in most places, values will be checked to be strictly equal to `undefined`, so it would be great to make this behavior consistent everywhere.

I'd guess several options to choose from:
- Take a value from the map and if it's undefined (`=== undefined`), replace it with the actual blinding.
- Take a value from the map and if it's nullish (`=== null` or `=== undefined`), replace it with the actual blinding.
- Check explicitly if some value was provided in the map using `blindings.has`, and if it is, use it independently of its value.

Seeing that most of the code already checks for the `undefined`, and in places where we allow optional args they will be initialized only in case of being `undefined`, I'd say it would be better to switch to `=== undefined` checks instead.

tests/scheme.ts
olegnn: It was declared only with a type definition without the actual implementation.

src/ps/proof.ts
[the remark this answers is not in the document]
olegnn: > And i thought you used void 0 over undefined because undefined is not a reserved word but now you are using undefined

I'm not using `undefined` in the code, I used `undefined` to distinguish `undefined` from `null` in the text. All other uses of `undefined` can be replaced too if needed.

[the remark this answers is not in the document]
lovesh: Ok. You can leave the existing `undefined`s in place.

PR #25, opened 2023-08-31
Boolean value support

cykoder:
Support for boolean values and boolean type in schema generation, encoded as positive integers. @lovesh let me know if i missed any area to test with this, or if there's a more appropriate encoder function.

Without boolean type support some credentials generate the schema with a boolean type as "string", which then fails schema validation on verification and ideally bools shouldn't be encoded as strings.

### DIFF
--- a/src/anonymous-credentials/schema.ts
+++ b/src/anonymous-credentials/schema.ts
@@ -982,7 +982,7 @@ export class CredentialSchema extends Versioned {
     return min >= 0 ? { type: this.POSITIVE_INT_TYPE } : { type: this.INT_TYPE, minimum: min };
   }
 
-  static parseBooleanType(): object {
+  static parseBooleanType(node: object, parsingOpts: ISchemaParsingOpts, nodeName: string): object {
     return { type: this.BOOLEAN_TYPE };
   }
 
--- a/src/anonymous-credentials/schema.ts
+++ b/src/anonymous-credentials/schema.ts
@@ -982,7 +982,7 @@ export class CredentialSchema extends Versioned {
     return min >= 0 ? { type: this.POSITIVE_INT_TYPE } : { type: this.INT_TYPE, minimum: min };
   }
 
-  static parseBooleanType(node: { minimum?: number }, parsingOpts: ISchemaParsingOpts, nodeName: string): object {
+  static parseBooleanType(): object {
     return { type: this.BOOLEAN_TYPE };
   }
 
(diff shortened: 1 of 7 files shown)


src/anonymous-credentials/schema.ts
```diff
@@ -972,6 +982,10 @@ export class CredentialSchema extends Versioned {
     return min >= 0 ? { type: this.POSITIVE_INT_TYPE } : { type: this.INT_TYPE, minimum: min };
   }
 
+  static parseBooleanType(node: { minimum?: number }, parsingOpts: ISchemaParsingOpts, nodeName: string): object {
```
lovesh, line 985: Remove `{ minimum?: number }`

src/anonymous-credentials/schema.ts
```diff
@@ -1054,6 +1071,8 @@ export class CredentialSchema extends Versioned {
         obj.properties[k] = CredentialSchema.getSubschema(v);
       }
       return obj;
+    } else if (typ === 'boolean') {
```
lovesh, line 1074: Better to move it to above check of `typ === 'string'` to reduce if-else branches

tests/anonymous-credentials/credential.spec.ts
```diff
@@ -152,6 +152,46 @@ describe(`${Scheme} Credential signing and verification`, () => {
...
+        isbool: { type: 'boolean' }
+      }
+    };
+    const credSchema = new CredentialSchema(schema);
+
+    const builder = new CredentialBuilder();
+    builder.schema = credSchema;
+
+    builder.subject = { fname: 'John', isnotbool: true };
+    expect(() => builder.sign(sk)).toThrow();
```
lovesh, line 170: Add one more check where subject has non boolean value but schema expects boolean, like `{isbool: "true"}`

lovesh (changes requested):

src/anonymous-credentials/schema.ts
cykoder: might be worth looking at integrating a linter into CI for easier catching of these, i know there is prettier but doesnt always give errors on unused fields etc. this was copy/pasted from other method to begin with.

i did want to ask though, do we need to supply a minimum at all? assuming not as it seems to work fine without it, maybe minimum of 0 makes sense if it actually changes anything?

src/anonymous-credentials/schema.ts
cykoder: this whole method can benefit from removing the "else" to make it a bit clearer as it always returns. updated for that

tests/anonymous-credentials/credential.spec.ts
cykoder: added check

PR #27, opened 2023-09-12
Bump crypto wasm to 0.21.0

nicobao:
See https://github.com/docknetwork/crypto-wasm/pull/13


lovesh:
Version 0.40.0 has been published with this change.

PR #28, opened 2023-09-14
Date and date-time format support

cykoder:
@lovesh lmk if i missed any spots, but i think this covers it. note: no auto generated schema test/impl because we cant reliably do that from JSON objects with date strings


src/anonymous-credentials/schema.ts
```diff
@@ -409,6 +409,7 @@ export class CredentialSchema extends Versioned {
   private static readonly INT_TYPE = 'integer';
   private static readonly POSITIVE_NUM_TYPE = 'positiveDecimalNumber';
   private static readonly NUM_TYPE = 'decimalNumber';
+  private static readonly DATE_TYPE = 'date';
```
lovesh, line 412: Should be datetime and not date. And bump schema version.

src/anonymous-credentials/schema.ts
```diff
@@ -626,6 +631,8 @@ export class CredentialSchema extends Versioned {
         return { type: ValueType.PositiveInteger };
       case CredentialSchema.BOOLEAN_TYPE:
         return { type: ValueType.PositiveInteger };
+      case CredentialSchema.DATE_TYPE:
+        return { type: ValueType.Integer, minimum: value['minimum'] };
```
lovesh, line 635: `ValueType` should be `PositiveInteger`. And you shouldn't need `minimum`.

src/anonymous-credentials/schema.ts
```diff
@@ -982,6 +992,11 @@ export class CredentialSchema extends Versioned {
     return min >= 0 ? { type: this.POSITIVE_INT_TYPE } : { type: this.INT_TYPE, minimum: min };
   }
 
+  static parseDateType(node: object, parsingOpts: ISchemaParsingOpts, nodeName: string): object {
```
lovesh, line 995: `nodeName` is unused

src/anonymous-credentials/schema.ts
```diff
@@ -982,6 +992,11 @@ export class CredentialSchema extends Versioned {
...
+    const min = parsingOpts.defaultMinimumInteger;
```
lovesh, line 996: Default minimum is negative but the numeric value of datetime won't ever be negative.  Are you planning to support some minimum hardocded date?
Secondly if the encoder is capable of converting datetime type to numeric minimum should not be numeric but a datetime.

src/encoder.ts
```diff
@@ -222,6 +222,30 @@ export class Encoder {
     };
   }
 
+  /**
+   * Returns an encoding function to be used on a message that is a date
+   */
+  static dateEncoder(minimum: number): EncodeFunc {
+    const f = Encoder.integerToPositiveInt(minimum);
+    return (v: unknown) => {
+      let dateVal = v instanceof Date ? v : null;
```
lovesh, line 231: if `v` is an instance of Date then next if block doesn't need to be considered. Just convert to int here.

src/encoder.ts
```diff
@@ -222,6 +222,30 @@ export class Encoder {
...
+      if (typeof v === 'string') {
+        const timestamp = Date.parse(v);
+        if (!Number.isNaN(timestamp)) {
+          dateVal = new Date(timestamp);
+        }
+      } else if (typeof v === 'number') {
```
lovesh, line 237: The intention was to support date/datetime strings only. Or it has changed?

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -242,11 +242,13 @@ export class PresentationBuilder extends Versioned {
...
     credIdx: number,
     attributeName: string,
-    min: number,
-    max: number,
+    vmin: number | Date,
+    vmax: number | Date,
     provingKeyId: string,
     provingKey?: LegoProvingKey | LegoProvingKeyUncompressed
   ) {
+    const min = vmin instanceof Date ? vmin.getTime() : vmin;
```
lovesh, line 250: The conversion of date to integer should be done with the encoder and not duplicated here.

lovesh (changes requested):

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -242,11 +242,13 @@ export class PresentationBuilder extends Versioned {
   enforceBounds(
     credIdx: number,
     attributeName: string,
-    min: number,
-    max: number,
+    vmin: number | Date,
```
lovesh, line 245: You are expecting `number`, `Date` and `string` in encoder so the same type should be used here.
And type would be changed in `BlindedCredentialRequestBuilder.enforceBounds` as well.

src/anonymous-credentials/schema.ts
maycon-mello: That's a good point. I was using a minimum placeholder in case it's not defined `new Date('0000-01-01')`, do you have any other suggestion? 

Actually, the timestamp for this placeholder minimum date is `-62167219200000` which is a negative number (not Unix timestamp)

src/anonymous-credentials/schema.ts
cykoder: incorrect, dates can be negative

src/anonymous-credentials/schema.ts
cykoder: >Default minimum is negative but the numeric value of datetime won't ever be negative

this is a wrong assumption. we want to support all dates, not just post 1970

src/anonymous-credentials/presentation-builder.ts
cykoder: what has this todo with the encoder? this is simplyifying the APi for the user so they can pass date objects instead of knowing to call getTime for enforceBounds

src/anonymous-credentials/presentation-builder.ts
cykoder: can you elaborate on `BlindedCredentialRequestBuilder`? what do you mean by the type will be changed?

> You are expecting number, Date and string in encoder so the same type should be used here.

but this conflicts with this feedback:

> The conversion of date to integer should be done with the encoder and not duplicated here.

src/anonymous-credentials/schema.ts
cykoder: i recall having some problems with the boolean encoder without having these arguments there, but i removed it for this + boolean and seems fine now. maybe im misremembering or something changed since

src/anonymous-credentials/presentation-builder.ts
cykoder: ignore previous comment about "BlindedCredentialRequestBuilder.enforceBounds" (deleted it) - i misread your comment. i can update this once we address the feedback for this method

src/anonymous-credentials/schema.ts
maycon-mello: I think we should support it, as it would be used for a birth date use case

lovesh: If you want to support dates older than 1970, then `-(Math.pow(2, 32) - 1)` is not a reasonable lower bound as that takes you about 50 days back only.

src/encoder.ts
lovesh: It's not about being strict or not what exactly are you aiming for.

src/anonymous-credentials/schema.ts
cykoder: agree, it wasnt so bad when unix epochs were used but we're now using js timestamps to be consistent. your input would be appreciated here on what a good lower bound would be, that would be supported by the encoder

src/anonymous-credentials/presentation-builder.ts
lovesh: I didn't say call `getTime` for the args. You're calling `getTime` directly here exposing encoder details when you already have access to the encoder inside this function and can call it on min, max. If in the future any change is made to the encode function, that change has to be made here as well.

src/anonymous-credentials/presentation-builder.ts
lovesh: > can you elaborate on BlindedCredentialRequestBuilder? what do you mean by the type will be changed?

Check the mentioned function. It's enforcing bounds as well.

> but this conflicts with this feedback:

I am talking about the type of function arg. In the encoder, you can/could pass the date as a `number|string|Date` but this expects only `number|Date`. Its inconsistent.

src/anonymous-credentials/presentation-builder.ts
cykoder: are you referring to `Encoder.dateEncoder`? that returns a function which returns an encoded message, rather than numeric value - so i dont think i want to use that directly. i could extract the string/date obj parsing into a utility called by both instead?

lovesh: Yes, I mean the function that converts to a number

src/anonymous-credentials/presentation-builder.ts
cykoder: >Check the mentioned function. It's enforcing bounds as well.

yeah i misread that in my original comment and amended it, assumnig you are replying to comments via email/app or similar and didnt see the update

> I am talking about the type of function arg. In the encoder, you can/could pass the date as a number|string|Date but this expects only number|Date. Its inconsistent.

im aware thats the suggestion here but maybe i was misunderstanding your point in the other thread, initially i thought you were saying *not* to do conversion in that method at all, but i think your point now is rather about sharing logic than actually calling the encoder method

src/anonymous-credentials/schema.ts
lovesh: You tell me as you're dealing with the app layer :). So you will need older dates for the age of people, buildings,  ancient artifacts, etc so the earliest dates being around the 14th century should be sufficient? `-(Math.pow(2, 44) - 1)` works for that. Make it a separate constant in `DefaultSchemaParsingOpts` called  `defaultMinimumDate` or something.

src/anonymous-credentials/presentation-builder.ts
lovesh: > yeah i misread that in my original comment and amended it, assumnig you are replying to comments via email/app or similar and didnt see the update

Sorry. Was replying through browse but didn't see that.

> your point now is rather about sharing logic than actually calling the encoder method

Yes.

src/anonymous-credentials/schema.ts
cykoder: if there are no library limitations as to how low the number can be, then sure i'll come up with something - no problem

src/anonymous-credentials/presentation-builder.ts
cykoder: makes sense now, thought you were referring to the actual encoding somehow and thought i misunderstood the code path from enforceBounds. will do that

src/anonymous-credentials/presentation-builder.ts
cykoder: got it now after resolving other thread, will update

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -242,11 +243,16 @@ export class PresentationBuilder extends Versioned {
...
     credIdx: number,
     attributeName: string,
-    min: number,
-    max: number,
+    vmin: number | Date | string,
+    vmax: number | Date | string,
     provingKeyId: string,
     provingKey?: LegoProvingKey | LegoProvingKeyUncompressed
   ) {
+    // TODO: This isn't clean because it's not checking if the attribute `attributeName` is of datetime type and then converting.
```
lovesh, line 251: This note is for me. We can take care of it in a later version.

tests/anonymous-credentials/credential.spec.ts
```diff
@@ -251,6 +251,92 @@ describe(`${Scheme} Credential signing and verification`, () => {
     });
   });
 
+  it('for credential with date-time fields', () => {
```
lovesh, line 254: Add some checks with older dates so to test negative timestamps.

src/anonymous-credentials/presentation-builder.ts
```diff
@@ -242,11 +243,16 @@ export class PresentationBuilder extends Versioned {
   enforceBounds(
     credIdx: number,
     attributeName: string,
-    min: number,
-    max: number,
+    vmin: number | Date | string,
```
lovesh, line 246: Update JSdoc

tests/anonymous-credentials/presentation.spec.ts
```diff
@@ -939,6 +966,65 @@ describe(`${Scheme} Presentation creation and verification`, () => {
...
+
+    const pkId = 'random';
+
+    // ------------------- Presentation with 1 credential -----------------------------------------
+    const builder7 = new PresentationBuilder();
+    expect(builder7.addCredential(credential7, pk1)).toEqual(0);
+
+    builder7.markAttributesRevealed(0, new Set<string>(['credentialSubject.name']));
+
+    const [minDateTime, maxDateTime] = [new Date('2022-09-14T14:26:40.488Z'), new Date('2024-09-14T14:26:40.488Z')];
```
lovesh, line 980: Do some checks where min corresponds to negative timestamps.

src/anonymous-credentials/schema.ts
cykoder: i think that your suggestion is a reasonable default, go too far and we may have problems - for any other use case before that date they can supply their own minimum value. i do wonder how "future proof" it JS timestamps are in this case, but we wont experience that issue for many years i think

lovesh (approved):

Issue #33, opened 2024-08-16
docs: need updates?


lanterno:
Hello, I'm trying to follow the documentation to implement the examples. However, the function names seem to have been modified from what I see in the docs. 

I managed to still get somewhere by trying my luck by looking for similarly named types. 

However right now, I'm trying to create a statement for my proof, but I can't find the `Statement.bbsPlusSignature` referenced in the docs. 

the section where it's referenced:
```
// Create a BBS+ signature, true indicates that attributes/messages are arbitrary bytes and should be encoded first
const statement1 = Statement.bbsPlusSignature(params, pk, revealedMsgs, true);
const statements = new Statements();
statements.add(statement1);
```

Can you help me find the right type I should use here? 

Finally, thanks for open-sourcing this library, it's really helpful (when I manage to make it work :) )

lovesh:
Thanks for reporting this and sorry for state of the docs. 

Instead of `Statement.bbsPlusSignature`, use `Statement.bbsPlusSignatureProverConstantTime` for the prover and `Statement.bbsPlusSignatureVerifierConstantTime` for the verifier. For more info, see [this](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/scheme.ts) where the statement for prover and verifier are set based on the scheme being used in testing. These are then used in these 2 utility functions, for [prover](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/util.ts#L17) and [verifier](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/util.ts#L30). Finally, a [test](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/scheme.spec.ts#L43) showing the usage.

For a more high level API, see [this](https://github.com/docknetwork/crypto-wasm-ts/tree/master/src/anonymous-credentials) and corresponding [tests](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/anonymous-credentials/presentation.spec.ts).

lanterno:
Hi @lovesh Thanks for your reply. very helpful. I'll try it today or tomorrow. 

I would be happy to do a PR to update the documentation after I finish my experiment with the library. Would that be helpful or not needed?

lovesh:
Hi @lanterno . The PR would be very helpful and welcome. And let me know of any other questions you have as well.

[...]

PR #35, opened 2024-10-27
Small fix in README.md

ineiti:
Actually not sure if this is correct - but the only other occurence of `sigPk` is a `BBSPlusPublicKeyG2`. At least it looks wrong to have the `sigPk` be of type `BBSPlusSecretKey`...


lovesh:
Yes, you're change is correct. 
Secondly, some of the functions shown in the readme have changed, see [this](https://github.com/docknetwork/crypto-wasm-ts/issues/33) for an example. Until this gets updated, have a look at [these tests](https://github.com/docknetwork/crypto-wasm-ts/tree/master/tests/composite-proofs).

lovesh (approved):

ineiti:
@lovesh #33 is from my colleague. I told him to look at how to improve the documentation a bit... Any pointers in what you think is the most important part?

We're using the library for one of our hands-on workshops, and the library is really great - incredible work! If we can contribute with some documentation, we'd be happy.

lovesh:
Thanks for your offer to help.
The conceptual documentation is fine. But some of the code examples are out of date so they need to be fixed by looking at the tests I mentioned before. Secondly there are some new features added like more protocols for range proofs which you can see in [these tests](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/bound-check.spec.ts) but missing in the Readme. 
The API described in the Readme is low-level and a higher level abstraction exists and is documented [here](https://github.com/docknetwork/crypto-wasm-ts/tree/master/src/anonymous-credentials) conceptually but lacks code examples which could be written referring [these tests](https://github.com/docknetwork/crypto-wasm-ts/tree/master/tests/anonymous-credentials), mostly [this one](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/anonymous-credentials/presentation.spec.ts). If you haven't seen this abstraction, I'll suggest taking a look. Its more ergonomic than the code examples in the main Readme. 
There is support for threshold BBS/BBS+ as seen in [this test](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/threshold-sigs.spec.ts) but missing from Readme.  
Finally, which parts have you had trouble using and think could benefit from more documentation, tests or a better API

ineiti:
Great - thanks a lot. I'll discuss it with @lanterno and come back to you. Is there already an issue regarding documentation?

lovesh:
Thanks. Just created [this issue](https://github.com/docknetwork/crypto-wasm-ts/issues/36). Feel free to ask any questions.

PR #37, opened 2024-11-12
Updates the documentation with BBS examples

lanterno:
This Pull Request aims to assist a little bit in the documentation by addressing some issues. 

Here's a list of all the done changes:
- Replace examples with others that follow specifically the BBS scheme. All the current examples reference generic function names as they exist in the tests. However, the tests work because they depend on `tests/scheme.spec.ts`, so  we.
- Removes `expect` JS function from jest.  The jest library is not easy to set up for typescript (if it's not already installed). Also, it's not required to test the library and function it.
- I reviewed in detail -and tested- the following sections
  - General Usage
  - Composite Proofs
  - Bound Checks using ZKP

I'm working on testing and improving the other sections, but I didn't want to make a single big PR, to make it easy to review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,28 +228,26 @@ the `encode` argument as true to encode it using your own encoding function.
   
   Letting the signing function encode  
   ```ts
-  // messages is a list of bytearrays and converted as mentioned above  
-  const messages: Uint8Array[] = [...];
+  import { BBSSignature } from '@docknetwork/crypto-wasm-ts';
   
   // The signing function will encode bytes to a field element as true is passed
-  const sig = SignatureG1.generate(messages, sk, params, true);
+  const sig = BBSSignature.generate(messages, sk, paramsDeterministc, true);
   
   // As the messages are not encoded, pass true to the verification function to make it encode messages before verifying the signature.
-  const result = sig.verify(messages, pk, params, true);
+  const result = sig.verify(messages, pk, paramsDeterministc, true);
   
   result.verified // true
   ```
   
   Passing pre-encoded messages to signing function
   ```ts
-  // messages is a list of bytearrays and converted as mentioned above  
-  const messages: Uint8Array[] = [...];
-  
+  const encodedMessages = [];
+
   for (let i = 0; i < messages.length; i++) {
     encodedMessages.push(generateFieldElementFromBytes(messages[i]));
   }
   // The signing function will not encode as false is passed
-  const sig = SignatureG1.generate(encodedMessages, sk, params, false);
+  const sig = BBSSignature.generate(encodedMessages, sk, params, false);
 
   // As the messages are pre-encoded, pass false to the verification function to avoid encoding messages before verifying the signature.
   const result = sig.verify(encodedMessages, pk, params, false);
@@ -258,9 +256,9 @@ the `encode` argument as true to encode it using your own encoding function.
 
 #### Proof of knowledge of signature
 
-Proving and verifying knowledge of signature can be done with or without using the composite proof system but this doc will only describe using the composite proof system. For the other way, see tests [here](./tests/bbs-plus.spec.ts)
+Proving and verifying knowledge of signature can be done with or without using the composite proof system but this doc will only describe using the composite proof system. For the other way, see tests [here](./tests/scheme.spec.ts)
 
-The code for BBS signature lives [here](./src/bbs-plus). 
+The code for BBS signature lives [here](./src/bbs/). 
 
 ### Accumulators
 
--- a/README.md
+++ b/README.md
@@ -764,8 +764,8 @@ witnesses.add(witness2);
 const proof = CompositeProof.generate(proofSpec, witnesses);
 ```
 
-##### Getting a blind signature
-
+##### Getting a blind signature (Example applies to BBS+)
+Disclaimer: With BBS, there is no blinding (the commitment is computationally hiding, though it can be made perfectly hiding by adding a dummy attribute). However, it can be easily achieved with BBS+.
 A complete example is shown in this [test](tests/composite-proofs/blind-signature.spec.ts).
 
 A signature is blind when the signer is not aware of the message (or a part of the message) that he is signing, thus the signer is blind.
@@ -780,7 +780,7 @@ prove that he knows that the commitment contains those 2 attributes
 // Messages are secret1, name, secret2, email, city
 
 // Signature params for 5 attributes
-const sigParams = BBSSignatureParams.generate(5, label);
+const sigParams = BBSPlusSignatureParamsG1.generate(5, label);
 
 // Prepare messages that will be blinded (hidden) and known to signer
 const blindedMessages = new Map();
@@ -797,10 +797,10 @@ The signature requester, prover in this case, creates a blind signature request.
 randomness `blinding` that goes into the commitment. This randomness is later used
 
 ```ts
-import { BBSBlindSignature } from '@docknetwork/crypto-wasm-ts';
+import { BBSPlusBlindSignatureG1 } from '@docknetwork/crypto-wasm-ts';
 
 // Blind signature request will contain a commitment, 
-const [blinding, request] = BBSBlindSignature.generateRequest(blindedMessages, params, true);
+const [blinding, request] = BBSPlusBlindSignatureG1.generateRequest(blindedMessages, params, true);
 ```
 
 The proof needs to be over only 1 `Statement`, the statement proving knowledge of the committed attributes in the commitment. 
@@ -819,13 +819,8 @@ const proofSpec = new ProofSpec(statements, new MetaStatements());
 
 Now the prover creates witness for the commitment `Statement` using the randomness and the hidden attributes.
 ```ts
-// The witness to the Pedersen commitment contains the blinding at index 0 by convention and then the hidden messages
-const committeds = [blinding];
-for (const i of blindedIndices) {
-  // The messages are encoded before committing
-  committeds.push(BBSSignature.encodeMessageForSigning(blindedMessages.get(i)));
-}
-const witness1 = Witness.pedersenCommitment(committeds);
+import { getBBSWitnessForBlindSigRequest } from '@docknetwork/crypto-wasm-ts';
+const witness1 = getBBSWitnessForBlindSigRequest(blindedMessages)
 const witnesses = new Witnesses();
 witnesses.add(witness1);
 
@@ -845,7 +840,7 @@ console.assert(proof.verify(proofSpec).verified);
 revealedMessages.set(1, stringToBytes('John Smith'));
 revealedMessages.set(3, stringToBytes('john.smith@emample.com'));
 revealedMessages.set(4, stringToBytes('New York'));
-const blindSig = BBSBlindSignature.generate(request.commitment, revealedMessages, sk, params, true);
+const blindSig = BBSPlusBlindSignatureG1.generate(request.commitment, revealedMessages, sk, params, true);
 ```
 
 The prover can now "unblind" the signature meaning he can convert a blind signature into a regular BBS signature 
--- a/README.md
+++ b/README.md
@@ -233,7 +233,8 @@ the `encode` argument as true to encode it using your own encoding function.
   
   // As the messages are not encoded, pass true to the verification function to make it encode messages before verifying the signature.
   const result = sig.verify(messages, pk, params, true);
-  expect(result.verified).toEqual(true);
+  
+  result.verified // true
   ```
   
   Passing pre-encoded messages to signing function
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,20 +76,20 @@ yarn test
 ## Overview
 Following is a conceptual explanation of the primitives.
 
-### BBS+ Signature
-BBS+ signature allow for signing an ordered list of messages, producing a signature of constant size independent of the number
+### BBS Signature
+BBS signature allow for signing an ordered list of messages, producing a signature of constant size independent of the number
```
ineiti, line 80: Please add a short disclaimer here about BBS/BBS+, similar to what is written in the [IRTF draft](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html), but shorter, and mentioning BBS and BBS+:

> The name BBS is derived from the authors of the original academic work by Dan Boneh, Xavier Boyen, and Hovav Shacham [[BBS04](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#BBS04)], where the scheme was first described as part of a group signatures protocol. Soon after, the scheme was described by Camenisch and Lysyanskaya as a stand-alone signatures scheme in [[CL04](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#CL04)], for anonymous credentials applications. Later, Au, Susilo an Mu presented the first, provably secure version of BBS Signatures in [[ASM06](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#ASM06)]. Following, works by Camenisch, Drijvers and Lehmann [[CDL16](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#CDL16)] and by Barki, Brunet, Desmoulins and Traore [[BBDT16](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#BBDT16)], proved the security of the scheme in settings where more efficient computations are possible, thereby improving performance. Finally, in 2023, Tessaro and Zhu, presented in [[TZ23](https://identity.foundation/bbs-signature/draft-irtf-cfrg-bbs-signatures.html#TZ23)] further performance improvements, shrinking the BBS signature. This document is mainly based on that work.

README.md
```diff
@@ -76,20 +76,20 @@ yarn test
...
 of messages. The signer needs to have a public-private keypair and signature parameters which are public values whose size
 depends on the number of messages being signed. A verifier who needs to verify the signature needs to know the
 signature parameters used to sign the messages and the public key of the signer. In the context of anonymous credentials, 
 messages are called attributes.  
-BBS+ signature also allow a user to request a blind signature from a signer where the signer does not know 1 or more messages
+BBS signature also allows a user to request a blind signature from a signer where the signer does not know 1 or more messages
```
ineiti, line 85: Either

A BBS signature also allows

or

BBS signatures also allow

README.md
```diff
@@ -177,83 +181,95 @@ Now each of the above list must be converted to bytearrays, i.e. `Uint8Array` an
 Before messages can be signed, 2 things are needed:
 
 - **Signature parameters**: Public values, that can be created by anyone but must be known to the signer and verifier to sign and verify respectively. To create them, the number of messages (attributes) being signed must be known and the size of the parameters increases with the number. In the above example, number of attributes is 4. These parameters can be generated randomly or deterministically by using a publicly known label. It is advised to use the latter as it allows for extending/shrinking the same parameters when number of messages change.
-- **Keypair**: To create and verify BBS+ signature, the signer (issuer in case of a credential) needs to create a secret key to sign, public key to verify. 
+- **Keypair**: To create and verify BBS signature, the signer (issuer in case of a credential) needs to create a secret key to sign, public key to verify.
```
ineiti, line 184: Same as above - but it seems that most other usages are `BBS signatures`, so I'd stick with the plural throughout the document, except where it doesn't make sense...

README.md
```diff
@@ -177,83 +181,95 @@ Now each of the above list must be converted to bytearrays, i.e. `Uint8Array` an
...
     encodedMessages.push(generateFieldElementFromBytes(messages[i]));
   }
   // The signing function will not encode as false is passed
-  const sig = SignatureG1.generate(encodedMessages, sk, params, false);
+  const sig = BBSSignature.generate(encodedMessages, sk, params, false);
 
   // As the messages are pre-encoded, pass false to the verification function to avoid encoding messages before verifying the signature.
   const result = sig.verify(encodedMessages, pk, params, false);
-  expect(result.verified).toEqual(true);
+  result.verified // true
```
ineiti, line 254: Either keep the `expect`, or do something like:

```
if (!result.verified){
  throw new Error("Signature verification failed");
}
```

README.md
```diff
@@ -177,83 +181,95 @@ Now each of the above list must be converted to bytearrays, i.e. `Uint8Array` an
...
   ```
 
 #### Proof of knowledge of signature
 
-Proving and verifying knowledge of signature can be done with or without using the composite proof system but this doc will only describe using the composite proof system. For the other way, see tests [here](./tests/bbs-plus.spec.ts)
+Proving and verifying knowledge of signature can be done with or without using the composite proof system but this doc will only describe using the composite proof system. For the other way, see tests [here](./tests/scheme.spec.ts)
 
-The code for BBS+ signature lives [here](./src/bbs-plus). 
+The code for BBS signature lives [here](./src/bbs/).
```
ineiti, line 261: plural

README.md
```diff
@@ -338,7 +354,7 @@ is the accumulator value when the witness was created, this value should be publ
   const verifAccumulator = PositiveAccumulator.fromAccumulated(accumulator.accumulated);
 
   // Note that only public values needed to verify the membership
-  expect(verifAccumulator.verifyMembershipWitness(e4, witness, pk, params)).toEqual(true);
+  verifAccumulator.verifyMembershipWitness(e4, witness, pk, params);  // true
```
ineiti, line 357: As above, keep the `expect`, or add a `throw`.

README.md
```diff
@@ -478,33 +495,39 @@ Both the prover and verifier should independently construct this `ProofSpec`. No
...
-const proof = CompositeProofG1.generate(proofSpec, witnesses, nonce);
+const proof = CompositeProof.generate(proofSpec, witnesses, nonce);
 ```
 
 Verifier can now verify this proof. Note that the verifier does not and must not receive `ProofSpec` from prover, it 
 needs to generate on its own.
 
 ```ts
-expect(proof.verify(proofSpec, nonce).verified).toEqual(true);
+proof.verify(proofSpec, nonce).verified;  // true
```
ineiti, line 527: `expect` or `throw`

README.md
```diff
@@ -787,32 +812,32 @@ Now the prover creates witness for the commitment `Statement` using the randomne
...
+const proof = CompositeProof.generate(proofSpec, witnesses);
 ```
 
 Signer now verifies the proof. Note that the signer independently creates the `ProofSpec` as he knows which attributes are being 
 hidden from him. If the proof is correct, signer creates a blind signature using the known attributes and the commitment 
 and sends to the prover.
 
 ```ts
-expect(proof.verify(proofSpec).verified).toEqual(true);
+proof.verify(proofSpec).verified;  // true
```
ineiti, line 829: `expect` or `throw`

README.md
```diff
@@ -830,7 +855,7 @@ for (const [i, m] of revealedMessages.entries()) {
 
 // Signature can be verified
 const result = sig.verify(messages, pk, params, true);
-expect(result.verified).toEqual(true);
+result.verified;  // true
```
ineiti, line 858: `expect` or `throw`

ineiti (commented):
Great - I put some small comments in there - mainly plural/singular for `BBS signatures` and verification of the signature should either `expect` or `throw` an error.

README.md
lovesh: Either "verify a BBS signature" or "verify BBS signatures".

README.md
lovesh: Even I prefer keeping an "error throwing" mechanism over a comment even for illustration purposes. Wdyt about `console.assert` over `expect`? The former is available  in browsers and node.js

README.md
```diff
@@ -763,8 +786,10 @@ The signature requester, prover in this case, creates a blind signature request.
 randomness `blinding` that goes into the commitment. This randomness is later used
```
lovesh, line 786: This is a place where BBS and BBS+ differ. With BBS, there is no `blinding` (the commitment is computationally hiding, though it can be made perfectly hiding by adding a dummy attribute). Can you please mention it here. And see comment below on `generateRequest`

README.md
```diff
@@ -763,8 +786,10 @@ The signature requester, prover in this case, creates a blind signature request.
...
 
 ```ts
+import { BBSBlindSignature } from '@docknetwork/crypto-wasm-ts';
+
 // Blind signature request will contain a commitment, 
-const [blinding, request] = BlindSignatureG1.generateRequest(blindedMessages, params, true);
+const [blinding, request] = BBSBlindSignature.generateRequest(blindedMessages, params, true);
```
lovesh, line 792: `BBSBlindSignature.generateRequest` returns a single item, i.e. `request` but `BBSPlusBlindSignatureG1.generateRequest` returns as above.

README.md
```diff
@@ -787,32 +812,32 @@ Now the prover creates witness for the commitment `Statement` using the randomne
 const committeds = [blinding];
```
lovesh, line 812: There are helpers for creating `Witness` object easy for blind signatures. See `getBBSWitnessForBlindSigRequest`, `getBBSPlusWitnessForBlindSigRequest`, `getPSWitnessesForBlindSigRequest`. See usage [here](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/social-kyc.spec.ts#L112).

README.md
```diff
@@ -787,32 +812,32 @@ Now the prover creates witness for the commitment `Statement` using the randomne
...
 // Signer creates a blind signature with these revealed messages and the commitment.
 revealedMessages.set(1, stringToBytes('John Smith'));
 revealedMessages.set(3, stringToBytes('john.smith@emample.com'));
 revealedMessages.set(4, stringToBytes('New York'));
-const blindSig = BlindSignatureG1.generate(request.commitment, revealedMessages, sk, params, true);
+const blindSig = BBSBlindSignature.generate(request.commitment, revealedMessages, sk, params, true);
 ```
 
-The prover can now "unblind" the signature meaning he can convert a blind signature into a regular BBS+ signature 
+The prover can now "unblind" the signature meaning he can convert a blind signature into a regular BBS signature
```
lovesh, line 840: Unblinding does not apply to BBS. See [this](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/social-kyc.spec.ts#L134).

lovesh (changes requested):
Mostly related to blind signartures

lovesh:
Thanks @lanterno for these fixes and @ineiti for your comments. Have suggested a few changes.

README.md
ineiti: I didn't know of `console.assert` - I like it. Yes, let's replace the `throw`/`expect` with `console.assert`.

lanterno:
@ineiti @lovesh  Hello, I've added many improvements on the language as mentioned above, and working on the other points, but I wanted to confirm one of the linguistic points I've changed in the last commit. 

That is:
many times "BBS signatures" is used but means "BBS signature scheme" while in other cases it really means "BBS signatures". So, I just to only use BBS when we mean the overarching scheme and use BBS signature/s otherwise.

So I started the documentation with this section 

> The BBS signature Scheme (henceforth referred to as just BBS)...

Then after use BBS whenever the scheme is meant.

README.md
lanterno: Sounds good! I used `console.assert(..);`

lovesh:
@lanterno 

> Then after use BBS whenever the scheme is meant. What do you think?

Sounds good.

lanterno:
@lovesh I updated the blinding part to use BBS+ and added a disclaimer at the top about the difference between BBS and BBS+ when it comes to blinding.  
This isn't perfect since we wanted to use only one crypto scheme through out all the examples, but it might be good to show the good side of BBS+ as well.

README.md
lanterno: I added a simple disclaimer at the top. Do you think there's something more to add there?

lovesh:
@lanterno Thank you!.

lanterno:
@lovesh Can you approve the workflow? (and maybe merge)

P.S. @ineiti can't approve PRs, he can only comment. (In case, you're waiting for this approval)

lovesh:
Thanks @lanterno .

Issue #36, opened 2024-10-29
Update docs


lovesh:
Some of the code samples in the Readme are obsolete. Also, some new features aren't documented in Readme. See [this](https://github.com/docknetwork/crypto-wasm-ts/issues/33) and [this](https://github.com/docknetwork/crypto-wasm-ts/pull/35).

ineiti:
From [your comment in #35](https://github.com/docknetwork/crypto-wasm-ts/pull/35#issuecomment-2441585931), we have the following points we can do easily:
- [ ] But some of the code examples are out of date so they need to be fixed by looking at the tests I mentioned before. 
- [ ] Secondly there are some new features added like more protocols for range proofs which you can see in [these tests](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/composite-proofs/msg-js-obj/bound-check.spec.ts) but missing in the Readme.
- [ ] The API described in the Readme is low-level and a higher level abstraction exists and is documented [here](https://github.com/docknetwork/crypto-wasm-ts/tree/master/src/anonymous-credentials) conceptually but lacks code examples which could be written referring [these tests](https://github.com/docknetwork/crypto-wasm-ts/tree/master/tests/anonymous-credentials), mostly [this one](https://github.com/docknetwork/crypto-wasm-ts/blob/master/tests/anonymous-credentials/presentation.spec.ts). If you haven't seen this abstraction, I'll suggest taking a look. Its more ergonomic than the code examples in the main Readme.

Regarding the BBS/BBS+ threshold part, we did not use that and wouldn't know how to correctly document it.

For your [last point](https://github.com/docknetwork/crypto-wasm-ts/pull/35#issuecomment-2441585931), where you ask us what would help us, I think the following would be great, what do you think? We would be happy to take on that part, too, after we finished with the above points.
1. Add an automatic API documentation generation like [typedoc](https://typedoc.org/)
2. Write the corresponding API documentation for the anonymous credentials
3. Re-write the README by distributing the current documentation throughout the code tree and making sure that the typedoc (or whatever we'll use) integrates these texts as well

lanterno:
Hey @lovesh, I actually discussed this with Linus, so I can respond since he seems busy.. 

Both 2. and 3. assume using a CLI tool that generates API interface documentation from the code rather than writing it manually. 

This morning, I was looking at https://typedoc.org/ (for the first time), and it generated a documentation just like this https://typedoc.org/example/ but with a less good UI. 

[...]

PR #39, opened 2024-11-25
Some minor improvements in the documentation and code examples

lanterno:
This PR adds some minor improvements to the documentation. 

But it changes the "BBS signature over varying number of messages" section significantly. Please review it, and let me know if I understand that feature correctly and renamed it well.

part of work towards https://github.com/docknetwork/crypto-wasm-ts/issues/36

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,17 +474,17 @@ const sig: BBSSignature = ...;
 // Prover prepares the attributes he wants to disclose, i.e. attribute index 2 and 4 (indexing is 0-based), and the ones he wants to hide. 
 const revealedMsgIndices: Set<number> = new Set();
 revealedMsgIndices.add(2);
-revealedMsgIndices.add(4);
+revealedMsgIndices.add(3);
 
 // revealedMsgs are the attributes disclosed to the verifier
 const revealedMsgs: Map<number, Uint8Array> = new Map();
 revealedMsgs.set(2, messages[2]);
+revealedMsgs.set(3, messages[3]);
 
 // unrevealedMsgs are the attributes hidden from the verifier
 const unrevealedMsgs: Map<number, Uint8Array> = new Map();
 unrevealedMsgs.set(0, messages[0]);
 unrevealedMsgs.set(1, messages[1]);
-unrevealedMsgs.set(3, messages[3]);
 ```
 
 Since there is only 1 kind of proof, i.e. the knowledge of a BBS signature and the signed attributes, there would be only 1 `Statement`. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -471,39 +471,40 @@ const pk: BBSPublicKey;
 // The signature
 const sig: BBSSignature = ...;
 
-// Prover prepares the attributes he wants to disclose, i.e. attribute index 2 and 4 (indexing is 0-based), and the ones he wants to hide. 
+// Prover prepares the attributes he wants to disclose, 
+// i.e. attribute index 2 and 4 (indexing is 0-based), and the ones he wants to hide. 
 const revealedMsgIndices: Set<number> = new Set();
 revealedMsgIndices.add(2);
-revealedMsgIndices.add(4);
+revealedMsgIndices.add(3);
```
lovesh, line 478: Why 4?

README.md
```diff
@@ -471,39 +471,40 @@ const pk: BBSPublicKey;
...
 
 // revealedMsgs are the attributes disclosed to the verifier
 const revealedMsgs: Map<number, Uint8Array> = new Map();
 revealedMsgs.set(2, messages[2]);
+revealedMsgs.set(3, messages[3]);
```
lovesh, line 483: Thanks. I missed it. But should be be 4.

README.md
```diff
@@ -28,7 +28,7 @@ the [WASM wrapper](https://github.com/docknetwork/crypto-wasm).
       - [Terminology](#terminology)
       - [Examples](#examples)
         - [Selective disclosure](#selective-disclosure)
-        - [BBS signature over varying number of messages](#bbs-signature-over-varying-number-of-messages)
+        - [BBS signature over null-valued messages](#bbs-signatures-over-null-valued-messages)
```
lovesh, line 31: Using null-values is just one way to deal with it as the example shows other way as well. That section is trying to solve the problem of varying number of messages.

lovesh (changes requested):

README.md
lanterno: I was trying to follow the messages array we constructed above which has only 4 messages. Now, by reviewing it, I see it's not suitable here, so I added a new messages array with 5 elements and changed the indexes accordingly. 

What do you think now?

README.md
lanterno: You're right. it's fixed now.

README.md
lovesh: Thanks.

README.md
lovesh: Thanks.

lovesh (approved):

PR #42, opened 2025-03-28
docs: adds small improvements to the accumulators section in documentation

lanterno:
Hello, as I was following through the documentation on the accumulator section, some things were confusing a little bit so I update them here. 

But one big problem was when I couldn't import `InMemoryState` from the library. Looking at the code, I found that it was not exported, so I fix this in the second commit in this PR by adding the line:
```ts
export * from './in-memory-persistence';
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Positive support only membership witnesses while universal support both membersh
 Creating non-membership witnesses is expensive however, and the cost depends on the number of members present in the accumulator.
 Both accumulators are owned by an accumulator manager who has the private key to the accumulator
 and only the owner can add or remove elements or create witnesses using the accumulator.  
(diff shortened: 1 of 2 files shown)


lovesh:
Hi. 

Regarding `in-memory-persistence`, that was just for testing so I didn't export it. Do you use it otherwise? 

Rest looks good.

lanterno:
@lovesh I needed the `in-memory-persistence` to try out the code examples in the documentation.

Is there another way to try this class during learning the framework without exporting it?

lovesh:
The `in-memory-persistence` documentation says its just for tests so its probably safe to export and expect people to be careful when using it. Merging the PR.

lovesh (approved):

